PR #9, opened 2022-03-18
Make the unique func id part of the definition instead of a separate arg

shapirone:
# Summary
Instead of accepting the function `id` as the first passed argument, we now have a `callback_id` param within the function's definition. 

## Why
This will allow us to eventually support adding a sensible default for the id, which would make it optional.
For more details, check out the [Slack Discussion](https://slack-pde.slack.com/archives/C02HWTPG26T/p1647558718035039?thread_ts=1647558718.035039&cid=C02HWTPG26T).

### Making `id` optional without this change
```ts
// Passing ID
DefineFunction("some_id", { ... definition });

// Default ID
DefineFunction(null, { ... definition });
```

### Making `id` optional with this change
```ts
// Passing ID
DefineFunction({ callback_id: "some_id",  ... definition });

// Default ID
DefineFunction({ ... definition });
```
## Testing
1. Check out this branch.
2. Pull down the template from the [neil-move-callbackid](https://github.com/slackapi/deno-reverse-string/tree/neil-move-callbackid) branch
3. Run `deno run -q --unstable --import-map=import_map.json --allow-read --allow-write --allow-net https://deno.land/x/deno_slack_builder@0.0.5/mod.ts`

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -33,7 +33,7 @@ Let's go over each of the arguments that must be provided to `DefineFunction`.
 
 #### Function Definition
 
-The second argument is the `definition` of the function, an object with a few properties that help to describe and define the function in more detail. In particular, the required properties of the object are:
+The first argument is the `definition` of the function, an object with a few properties that help to describe and define the function in more detail. In particular, the required properties of the object are:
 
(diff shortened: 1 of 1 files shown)


mwbrooks:
Nice @shapirone!

PR #10, opened 2022-04-08
Remove client

shapirone:
# Summary
Remove client from the SDK since we now have the client in a separate repo


docs/functions.md
```diff
@@ -57,9 +57,6 @@ The single argument to your function is an object composed of several properties
 
 - `env`: represents environment variables available to your function's execution context.
 - `inputs`: an object containing the input parameters you defined as part of your Function Definition. In the example above, the `name` input parameter is available on the `inputs` property of our function handler context.
-- `client`: a pre-configured instance of a Slack API Client. The API client has an async `call` function which accepts two arguments:
```
filmaj, line 60: I think we should be explicit about the `token` argument available now, though, yeah?

filmaj (approved):

docs/functions.md
mwbrooks: Yea, I agree. If we need to pass the `token` into the imported client then we should documentation about getting the token. Would that go into the [Function context docs?](https://github.com/slackapi/deno-slack-sdk/blob/eb8a42dc53a6e5a4d7ee95bab9b5b7eb8d03a639/docs/functions.md#function-handler-context)

mwbrooks (approved):

docs/functions.md
shapirone: Agree — I need to add `token` to the `FunctionContext` type, so I can address the documentation portion as part of that PR

PR #12, opened 2022-04-11
Initial support for Slack App Datastore

chris-skud:
Initial minimal datastore support allowing for inclusion in the app manifest.

Example definition:
```
import { DefineDatastore, Schema } from "slack-cloud-sdk/mod.ts";

export const MyDatastore = DefineDatastore("my_datastore", {
  primary_key: "id_attr",
  attributes: {
    id_attr: {
      type: Schema.types.string,
    },
    attr2: {
      type: Schema.types.string,
    },
    attr3: {
      type: Schema.types.int,
    },
  },
});

```
Then MyDatastore can be added to the manifest's `datastores` property.


shapirone (changes requested):
This is looking great 🎉  IMO we should address my other comments before merging:
- Using the `object` type in a test
- Moving the `name` arg into the `definition` as a key

Some additional fast-follow changes identified for after this is merged (mostly for my own note-keeping):
- We'll want to auto-register any `type` referenced by a `datastore`
- We'll want to auto-append `datastore:read` and `datastore:write` to `botScopes` if we find that the app has at least one `datastore`
- As called out in [a TODO](https://github.com/slackapi/deno-slack-sdk/pull/12/files#diff-11a92d799f9197d820de409d3502b5a25e4be3459f007e766a6c13c12d03d71fR24), we'll want to type what is passed to `type`. Right now in [this test](https://github.com/slackapi/deno-slack-sdk/pull/12/files#diff-49d3e0b3cce5facdf796b3b036050458eed0e13caf0666a876b49ec08b646d21R26) we're passing `"int"` which is invalid and we should catch

PR #13, opened 2022-04-12
Change the supported types for reignite

shapirone:
# Summary
Removing types that are not going to be supported at re-ignite.

Since the majority of these types will be re-added, I've commented them out for the time being.

## Justification
For re-ignite we are only supporting the following types, based on [this doc](https://corp.quip.com/XaUHABd6ZPUp/Reduce-function-input-type-support-to-match-Block-Kit-Decision):
- `string`
- `boolean`
- `user`
- `channel`
- `array` of
  - `string` as an `enum`
  - `user`
  - `channel

### DIFF
--- a/src/parameters/mod.ts
+++ b/src/parameters/mod.ts
@@ -1,12 +1,12 @@
-import SchemaTypes from "../schema/schema_types.ts";
+// import SchemaTypes from "../schema/schema_types.ts";
 import type {
   CustomTypeParameterDefinition,
-  TypedObjectParameterDefinition,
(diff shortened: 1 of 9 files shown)


src/parameters/mod.ts
```diff
@@ -1,12 +1,12 @@
-import SchemaTypes from "../schema/schema_types.ts";
+// import SchemaTypes from "../schema/schema_types.ts";
```
shapirone, line 1: We could likely remove this file (or the majority of it) since `ParameterVariable` isn't used without workflows

PR #20, opened 2022-04-13
Make function handler types optional

shapirone:
# Summary

These args are always passed in by the runtime, but having the requirement for `Env`, `token`, and `event` args means that they must be provided via tests. I think we can loosen this up as optional params rather than require them in to be defined in every test file.

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -38,12 +38,12 @@ export type FunctionHandlerReturnArgs<OutputParameters> = {
 
 export type FunctionContext<InputParameters> = {
   /** A map of string keys to string values containing any environment variables available and provided to your function handler's execution context. */
-  env: Env;
+  env?: Env;
   /** The inputs to the function as defined by your function definition. */
   // TODO: Support types generated from manifest
   inputs: InputParameters;
-  token: string;
-  event: FunctionInvocationBody["event"];
+  token?: string;
+  event?: FunctionInvocationBody["event"];
 };
 export interface ISlackFunction<
   InputParameters extends ParameterSetDefinition,


sopokuagyemang (approved):

selfcontained:
What if we provided a function that helped with building some things like function context arguments?

```ts
import { assertEquals } from "https://deno.land/std@0.99.0/testing/asserts.ts";
import { SlackFunctionTester } from "deno-slack-sdk/mod.ts";
import ReverseString from "./reverse.ts";

const { createContext } = SlackFunctionTester("reverse_string");

  const inputs = { stringToReverse: "foo" };
  const { outputs } = await ReverseString(createContext({ inputs }));
});
```

Maybe initially it's just a `createContext` helper it provides, but perhaps there's other things we could do with it, or use it as a way to allow devs to provide defaults for the context created. It might look something like this:

```ts
export type SlackFunctionTesterArgs<InputParameters> = {
  inputs: InputParameters;
  env?: Env;
  token?: string;
  event?: FunctionInvocationBody["event"];
};

export const SlackFunctionTester = (callbackId: string) => {
  const now = new Date();
  const testFnID = `fn${now.getTime()}`;

  const createContext = <InputParameters>(
    args: SlackFunctionTesterArgs<InputParameters>,
  ): FunctionContext<InputParameters> => {
    const ts = new Date();

    return {
      inputs: args.inputs,
      env: args.env || {},
      token: args.token || "slack-function-test-token",
      event: args.event || {
        type: "function_executed",
        event_ts: `${ts.getTime()}`,
        function_execution_id: `fx${ts.getTime()}`,
        inputs: args.inputs as Record<string, unknown>,
        function: {
          id: testFnID,
          callback_id: callbackId,
          title: "Function Test Title",
        },
      },
    };
  };

  return { createContext };
};
```

shapirone:
@selfcontained put together a better solution in https://github.com/slackapi/deno-slack-sdk/pull/22. Closing this PR! 🙌

PR #21, opened 2022-04-14
Enforce stricter assertions

lucasmichot:
This PR enforces stricter assertions, as very often comparison are done between primitives


selfcontained (approved):
Thanks, certainly seems like a better default assertion for where we're using it. Appreciate all of the improvements you've been contributing 😄

PR #22, opened 2022-04-14
adding a SlackFunctionTester to help with writing function tests

selfcontained:
## Summary
This creates a new top level export of `SlackFunctionTester` that can be used to ease writing tests for Slack Functions. Consuming it would like like the following when writing a test for  a function in your project:

```ts
// Can call this anywhere in your test file
const { createContext } = SlackFunctionTester("reverse_string");
```

And then in your tests where you need to pass arguments into your function...
```ts
const { outputs } = await ReverseString(createContext({ inputs }));


A full test would like like the following:
```ts
import { assertEquals } from "https://deno.land/std@0.99.0/testing/asserts.ts";
import ReverseString from "./reverse.ts";
import { SlackFunctionTester } from "deno-slack-sdk/mod.ts";

const { createContext } = SlackFunctionTester("reverse_string");

  const inputs = { stringToReverse: "foo" };
  const { outputs } = await ReverseString(createContext({ inputs }));
});
```

I think we could expand on what helper functions `SlackFunctionTester()` returns if we need over time, but initially this will help developers have more realistic arguments into their functions w/ low effort.

### DIFF
--- a/src/functions/function_tester.ts
+++ b/src/functions/function_tester.ts
@@ -1,12 +1,9 @@
-import { Env } from "../types.ts";
-import { FunctionContext, FunctionInvocationBody } from "./types.ts";
-
-type SlackFunctionTesterArgs<InputParameters> = {
-  inputs: InputParameters;
-  env?: Env;
-  token?: string;
-  event?: FunctionInvocationBody["event"];
-};
+import type { FunctionContext } from "./types.ts";
+type SlackFunctionTesterArgs<InputParameters> =
+  & Partial<FunctionContext<InputParameters>>
+  & {
+    inputs: InputParameters;
+  };
 
 export const SlackFunctionTester = (callbackId: string) => {
   const now = new Date();
--- a/src/functions/function_tester.ts
+++ b/src/functions/function_tester.ts
@@ -1,0 +1,39 @@
+import { Env } from "../types.ts";
+import { FunctionContext, FunctionInvocationBody } from "./types.ts";
+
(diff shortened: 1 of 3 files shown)


selfcontained:
Assuming we are ok with this approach, and after any changes, I can go ahead and get the default project template test updated to use this.

src/functions/function_tester.ts
```diff
@@ -0,0 +1,39 @@
+import { Env } from "../types.ts";
+import { FunctionContext, FunctionInvocationBody } from "./types.ts";
+
+type SlackFunctionTesterArgs<InputParameters> = {
+  inputs: InputParameters;
+  env?: Env;
+  token?: string;
+  event?: FunctionInvocationBody["event"];
+};
```
shapirone, line 9: nit: I'm thinking that we should tie this a bit closer to the `FunctionContext` itself and we could leverage a `Partial` to make things optional. I'm thinking something like

```suggestion
import type { FunctionContext } from "./types.ts";

type SlackFunctionTesterArgs<InputParameters> =
  & Partial<FunctionContext<InputParameters>>
  & {
    inputs: InputParameters;
  };
```

shapirone (approved):
This LGTM and sets up a nice precedent for test helpers if we ever need to provide more 🙌 

Just one optional nit suggestion!

src/functions/function_tester.ts
selfcontained: Good call, helps it stay in sync w/ that type.

shapirone (approved):
:chefs-kiss:

PR #24, opened 2022-04-14
[HERMES-2473] Function return args no longer require an output if there is an error

shapirone:
JIRA TIcket: [HERMES-2473](https://jira.tinyspeck.com/browse/HERMES-2473)

### Summary

If we catch an error in the function execution, we should be able to return the error without needing to set any outputs.

### Description
Right now if we return from a caught error we need to provide outputs
```ts
  try {
    throw new Error();
  } catch (e) {
    return await {
      error: e,
      outputs: {}
    };
  }
  ```
  
  but this isn't required by our BE so we shouldn't require `outputs` in the response if `error` is provided
  ```ts
```ts
  try {
    throw new Error();
  } catch (e) {
    return await {
      error: e,
    };
  }
  ```

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -30,11 +30,23 @@ export type FunctionHandler<InputParameters, OutputParameters> = {
   ): Promise<FunctionHandlerReturnArgs<OutputParameters>>;
 };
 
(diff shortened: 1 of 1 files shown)


selfcontained (approved):
Nice, tested this out and works as expected.

PR #25, opened 2022-04-14
Move the required field under the properties

shapirone:
### Summary
Moving the required field beneath the properties

### Details
There was confusion about a TS error around a required field that wasn't actually a parameter yet. We haven't figured out the right way to give specific error messaging for this kind of situation with our types, but it makes more sense to define the properties that are available before we say which are required.

Convo: https://slack-pde.slack.com/archives/C03BS52QLPJ/p1649958372816539
Relevant PR: https://github.com/slackapi/deno-reverse-string/pull/24

### DIFF
--- a/src/functions/functions_test.ts
+++ b/src/functions/functions_test.ts
@@ -25,7 +25,6 @@ Deno.test("Function with required params", () => {
     title: "All Types Function",
     source_file: "functions/example.ts",
     input_parameters: {
-      required: ["myString" /* , "myNumber" */],
       properties: {
         myString: {
           type: Schema.types.string,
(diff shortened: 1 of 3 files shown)


selfcontained (approved):
Changes seem fine, I'm not sure I follow what the advantage is to moving `required` below `properties` in all of these places. Certainly seems fine though. I can see how shifting it in our app code, like the reverse string app, is helpful in the clarity it provides though. Is the change here mostly just for consistency?

shapirone:
> Changes seem fine, I'm not sure I follow what the advantage is to moving `required` below `properties` in all of these places. Certainly seems fine though. I can see how shifting it in our app code, like the reverse string app, is helpful in the clarity it provides though. Is the change here mostly just for consistency?

💯 just for consistency and we could have gotten away without these changes to the SDK

PR #28, opened 2022-04-15
Support async and sync function handlers

selfcontained:
## Summary

This goes along with the following PR for the runtime: https://github.com/slackapi/deno-slack-runtime/pull/12

After that runtime is released, we can release this type change to allow sync or async function handlers. This will allow someone to write a function handler w/o the `async` keyword if they don't need it. So instead of:

```ts
const reverse: FunctionHandler<any, any> = async ({ inputs, env }) => {
  const reverseString = inputs.stringToReverse.split("").reverse().join("");

  return await {
    outputs: { reverseString },
  };
};
```

It could just be:

```ts
const reverse: FunctionHandler<any, any> = ({ inputs, env }) => {
  const reverseString = inputs.stringToReverse.split("").reverse().join("");

  return {
    outputs: { reverseString },
  };
};
```

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -24,11 +24,21 @@ export type FunctionInvocationBody = {
   };
 };
 
-export type FunctionHandler<InputParameters, OutputParameters> = {
+type AsyncFunctionHandler<InputParameters, OutputParameters> = {
   (
     context: FunctionContext<InputParameters>,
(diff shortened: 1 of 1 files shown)


shapirone (approved):
LGTM!

shapirone:
Just noting for future us — with the above nit we talked about later introducing a `FunctionHandlerSync` type so that there isn't always this requirement

PR #32, opened 2022-04-25
Updating datastore types with generics and making definition public

selfcontained:
## Summary

Making the `definition` property of a Datastore instance public (so we can reference it [for typed apis](https://github.com/slackapi/deno-slack-api/pull/9)). Also added generics for `name` and `primary_key` to support stricter typing on apis using the definitions as generics.

## Testing
This should change anything with how a datastore is created/referenced in the SDK alone, but will allow testing a new PR to the api client w/ typed datastore methods.

https://github.com/slackapi/deno-slack-api/pull/9

### DIFF
--- a/src/datastore/mod.ts
+++ b/src/datastore/mod.ts
@@ -8,20 +8,29 @@ import {
 
 /**
  * Define a datastore and primary key and attributes for use in a Slack application.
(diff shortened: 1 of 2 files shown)


shapirone (approved):
LGTM! Minor nit on dependencies on the API PR

PR #34, opened 2022-04-29
SlackFunctionTester now also accepts Function definitions

shapirone:
# Summary
`SlackFunctionTester` now accepts Function Definitions

## Details
This adds an overload to `SlackFunctionTester` where we now allow a value returned from `DefineFunction` to be passed into `SlackFunctionTester`. We'll still be able to pass `callbackId` strings as well, and any instance where you pass a callbackId (rather than a Function Definition) into a typed handler should still support typeahead within `SlackFunctionTester.CreateContext` based on the handler.

Both of the following scenarios should work w/ typeahead:

### Passing Function Definition

```ts
/* manifest.ts */
export const ReverseFunction = DefineFunction({
  callback_id: "reverse",
  title: "Reverse",
  description: "Takes a string and reverses it",
  source_file: "functions/reverse.ts",
  input_parameters: {
    properties: {
      stringToReverse: {
        type: Schema.types.string,
        description: "The string to reverse",
      },
    },
    required: ["stringToReverse"],
  },
  output_parameters: {
    properties: {
      reverseString: {
        type: Schema.types.string,
        description: "The string in reverse",
      },
    },
    required: ["reverseString"],
  },
});

/* functions/reverse_test.ts */
import { ReverseFunction } from "../manifest.ts";

const { createContext } = SlackFunctionTester(ReverseFunction);

  const { outputs } = await ReverseString(
    createContext({ inputs: { stringToReverse: "foo" } }), // ahoy type ahead!
  );
});

```

### Passing Callback ID

```ts
const { createContext } = SlackFunctionTester("reverse");

  const { outputs } = await ReverseString(
    createContext({ inputs: { stringToReverse: "foo" } }), // ahoy type ahead!
  );
});
```


src/functions/function_tester.ts
```diff
@@ -29,7 +52,7 @@ export const SlackFunctionTester = (callbackId: string) => {
         inputs: args.inputs as Record<string, unknown>,
         function: {
           id: testFnID,
-          callback_id: callbackId,
+          callback_id: func.definition.callback_id,
           title: "Function Test Title",
```
selfcontained, line 56: we should grab this from the `func.definition.title`

src/functions/tester/types.ts
```diff
@@ -0,0 +1,64 @@
...
+    InputParameters,
+    RequiredInput
+  >;
+
+  // Accept a string
+  (funcOrCallbackId: string): {
+    createContext: {
+      <I>(args: SlackFunctionTesterArgs<I>): FunctionContext<I>;
+    };
+  };
```
shapirone, line 63: This is the trickiest bit of this PR. Using `SlackFunctionTesterResponse` like above I would run into a problem when I tried to use a handler typed to our `BaseSlackFunctionHandler` and pass the result of `createContext` into that handler.

This was the best way I found to infer the type passed to a `createContext` call that worked for both `BaseSlackFunctionHandler` and `SlackFunctionHandler`. Open to any ideas on how this could be improved

src/functions/slack_function_handler_type_test.ts
```diff
@@ -37,10 +37,14 @@ Deno.test("SlackFunctionHandler with inputs and outputs", () => {
...
-  const { createContext } = SlackFunctionTester("test");
+  const { createContext } = SlackFunctionTester(TestFn);
   const inputs = { in: "test" };
   const result = handler(createContext({ inputs }));
-  assertEquals(result.outputs?.out, inputs.in);
+
+  const stringTest = SlackFunctionTester("test");
+  const stringResult = handler(stringTest.createContext({ inputs }));
+
+  assertEquals(result.outputs?.out, inputs.in, stringResult.outputs?.out);
```
shapirone, line 47: If a developer uses `SlackFunctionHandler` we expect them to pass that same definition to `SlackFunctionTester`, but just wanted to add a quick sanity test to make sure that just in case they don't, it still works.

src/functions/tester/mod.ts
```diff
@@ -0,0 +1,55 @@
...
+      OutputParameters,
+      RequiredInput,
+      RequiredOutput
+    >,
+) => {
+  const now = new Date();
+  const testFnID = `fn${now.getTime()}`;
+  const createContext: CreateContext<InputParameters, RequiredInput> = (
+    args,
+  ) => {
```
shapirone, line 28: Updated the createContext typing here

src/functions/tester/mod.ts
```diff
@@ -0,0 +1,55 @@
...
+  RequiredOutput extends PossibleParameterKeys<OutputParameters>,
+>(
+  funcOrCallbackId:
+    | string
+    | SlackFunction<
+      InputParameters,
+      OutputParameters,
+      RequiredInput,
+      RequiredOutput
+    >,
```
shapirone, line 22: Allowed the `string` type

src/functions/tester/mod.ts
```diff
@@ -0,0 +1,55 @@
...
+      event: args.event || {
+        type: "function_executed",
+        event_ts: `${ts.getTime()}`,
+        function_execution_id: `fx${ts.getTime()}`,
+        inputs: args.inputs as Record<string, unknown>,
+        function: {
+          id: testFnID,
+          callback_id: typeof funcOrCallbackId === "string"
+            ? funcOrCallbackId
+            : funcOrCallbackId.definition.callback_id,
```
shapirone, line 47: Added this type guard (it's the only place logic changes)

src/functions/tester/mod.ts
```diff
@@ -0,0 +1,55 @@
+import {
```
shapirone, line 1: Moved `SlackFunctionTester` into it's own folder, but this file is generally the same as before — going to call out the big changes so it's easier to review.

src/functions/tester/types.ts
```diff
@@ -0,0 +1,64 @@
...
+  RequiredInput extends PossibleParameterKeys<InputParameters>,
+> = {
+  (
+    args: SlackFunctionTesterArgs<
+      FunctionRuntimeParameters<InputParameters, RequiredInput> | undefined
+    >,
+  ): FunctionContext<
+    FunctionRuntimeParameters<InputParameters, RequiredInput>
+  >;
+};
```
shapirone, line 29: This `CreateContext` type is new, but unfortunately couldn't figure out how to re-use it

src/functions/tester/types.ts
selfcontained: Umm, i wouldn't have even come up with this idea, so def. don't have other ideas. It works amazingly though!

selfcontained (approved):
Nice job.

PR #35, opened 2022-05-26
Workflow Definition support

selfcontained:
## Adding ability to define workflows in your manifest

This exports a `DefineWorkflow()` from the main `mod.ts` file. It will create a workflow definition to be inserted into the `workflows` array in the manifest.

_Example of workflow being defined_
```ts
const ReacjiChannelerWorkflow = DefineWorkflow({
  callback_id: "reacji_channeler",
  title: "Reacji Channeler",
  input_parameters: {
    properties: {
      forward_channel: {
        type: Schema.slack.types.channel_id,
      },
      message_channel: {
        type: Schema.slack.types.channel_id,
      },
      message_ts: {
        type: Schema.types.string,
      },
    },
    required: ["forward_channel", "message_channel", "message_ts"],
  },
});
```

A workflow isn't very fun (or valid) without some steps. Steps can be added by one of two ways:

## 1️⃣ Calling with a typed function definition
This is what you have after calling `DefineFunction()`

```ts
const GetMessagePermalink = DefineFunction(...);

const ReacjiChannelerWorkflow = DefineWorkflow(...);

const step1 = ReacjiChannelerWorkflow.addStep(GetMessagePermalink, {
  channel: ReacjiChannelerWorkflow.inputs.message_channel,
  message_ts: ReacjiChannelerWorkflow.inputs.message_ts,
});
```

## 2️⃣ Using plain function references
You can also add a step to a workflow with the plain function reference string and inputs. This is helpful if you don't have the function definition available in your code. The only disadvantage here is you have to create the function reference string yourself, and your input values object isn't typed for you. Building off our `ReacjiChannelerWorkflow` example, we can add a slack builtin function as a step like so:

```ts
ReacjiChannelerWorkflow.addStep("slack#/functions/send_message", {
  channel_id: ReacjiChannelerWorkflow.inputs.forward_channel,
  message: `Link: ${step1.outputs.permalink}`,
});
```

## Adding your workflow to your manifest

```ts
export default Manifest({
  //...
  workflows: [ReacjiChannelerWorkflow],
});
```

## Testing
* I've added a workflow-level unit test to exercise defining a workflow and adding steps both ways.
* Define and add a workflow to your manifest, then run `hermes manifest` and check your manifest for your new workflow definition.
* You can deploy your app with this `workflow` in your manifest, which does accept a valid workflow, but we aren't quite ready to test running a workflow defined in the manifest. I would consider the manifest updating with your workflow as a success for now.

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -1,6 +1,7 @@
 import { Env, ManifestFunctionSchema } from "../types.ts";
 import {
   ParameterDefinition,
+  ParameterPropertiesDefinition,
   ParameterSetDefinition,
   RequiredParameters,
 } from "../parameters/mod.ts";
@@ -177,7 +178,7 @@ export interface ISlackFunction<
 export type FunctionDefinitionArgs<
   InputParameters extends ParameterSetDefinition,
   OutputParameters extends ParameterSetDefinition,
-  RequiredInput extends RequiredParameters<InputParameters>,
+  RequiredInputs extends RequiredParameters<InputParameters>,
   RequiredOutputs extends RequiredParameters<OutputParameters>,
 > = {
   callback_id: string;
@@ -187,13 +188,13 @@ export type FunctionDefinitionArgs<
   /** An optional description for your function. */
   description?: string;
   /** An optional map of input parameter names containing information about their type, title, description, required and (additional) properties. */
-  "input_parameters"?: {
-    properties: InputParameters;
-    required: RequiredInput;
-  };
+  "input_parameters"?: ParameterPropertiesDefinition<
+    InputParameters,
+    RequiredInputs
+  >;
   /** An optional map of output parameter names containing information about their type, title, description, required and (additional) properties. */
-  "output_parameters"?: {
-    properties: OutputParameters;
-    required: RequiredOutputs;
-  };
+  "output_parameters"?: ParameterPropertiesDefinition<
+    OutputParameters,
+    RequiredOutputs
+  >;
 };
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -3,7 +3,7 @@ import {
   ParameterDefinition,
   ParameterPropertiesDefinition,
   ParameterSetDefinition,
-  RequiredParameters,
+  PossibleParameterKeys,
 } from "../parameters/mod.ts";
 import { TypedArrayParameterDefinition } from "../parameters/types.ts";
 import type SchemaTypes from "../schema/schema_types.ts";
@@ -65,7 +65,7 @@ type TypedArrayFunctionInputRuntimeType<
  */
 type FunctionRuntimeParameters<
   Params extends ParameterSetDefinition,
-  RequiredParams extends RequiredParameters<Params>,
+  RequiredParams extends PossibleParameterKeys<Params>,
 > =
   & {
     [k in RequiredParams[number]]: FunctionInputRuntimeType<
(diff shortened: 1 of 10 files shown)


src/functions/types.ts
```diff
@@ -187,13 +188,13 @@ export type FunctionDefinitionArgs<
   /** An optional description for your function. */
   description?: string;
   /** An optional map of input parameter names containing information about their type, title, description, required and (additional) properties. */
-  "input_parameters"?: {
-    properties: InputParameters;
-    required: RequiredInput;
-  };
+  "input_parameters"?: ParameterPropertiesDefinition<
```
selfcontained, line 191: I found a place to re-use this existing Type we had, so wanted to make things a bit more consistent. I did test that this still works for function-related code.

src/parameters/mod.ts
```diff
@@ -124,16 +125,16 @@ export const ParameterVariable = <P extends ParameterDefinition>(
...
-//   namespace: string,
-//   paramName: string,
-// ): UntypedObjectParameterVariableType => {
-//   return WithUntypedObjectProxy(
-//     {},
-//     namespace,
-//     paramName,
-//   ) as UntypedObjectParameterVariableType;
-// };
+export const CreateUntypedObjectParameterVariable = (
```
selfcontained, line 128: I'm using this as an untyped variable object for `step.outputs` if you call workflow.addStep() with a plain function reference string.

src/workflows/types.ts
```diff
@@ -0,0 +1,83 @@
...
+    [name in Required[number]]: ParameterVariableType<
+      Parameters[name]
+    >;
+  }
+  & {
+    [name in keyof Parameters]?: ParameterVariableType<Parameters[name]>;
+  };
+
+// Workflow Step inputs are different than workflow inputs/outputs or workflow step outputs.
+// They are purley the config values for the step, and not definitions that can be referenced
```
shapirone, line 70:
```suggestion
// They are purely the config values for the step, and not definitions that can be referenced
```

src/manifest.ts
```diff
@@ -69,6 +69,10 @@ export class SlackManifest {
   }
 
   private registerFeatures() {
+    this.definition.workflows?.forEach((workflow) => {
+      workflow.registerStepFunctions(this);
+      workflow.registerParameterTypes(this);
+    });
```
shapirone, line 75: I noticed that running `hermes manifest` will register the functions I use in the workflow but the workflow itself isn't being added. I think this is because we seem to be missing the logic in the manifest export method to take the workflows from the manifest definition and add it to the manifest we return. Should look something like
```ts
if (def.workflows) {
      manifest.workflows = def.workflows?.reduce((acc = {}, fn) => {
        acc[fn.id] = fn.export();
        return acc;
      }, {} as ManifestSchema["workflows"]);
    }
```

shapirone (changes requested):
This is looking pretty good!

The big concern is that I tested this via `hermes manifest` but I didn't see any `workflows` in the manifest. I did see the functions and types I referenced get auto-registered though! We should also add some Manifest level tests to make sure we're successfully auto-registering types and functions used by workflows.

Some smaller concerns about some typing and the names of some types.

src/parameters/mod.ts
```diff
@@ -21,9 +21,10 @@ export type RequiredParameters<
 
 export type ParameterPropertiesDefinition<
   Parameters extends ParameterSetDefinition,
+  Required extends RequiredParameters<Parameters>,
```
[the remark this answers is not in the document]
selfcontained: Yah... I would not be surprised if there's a better way to do this...

shapirone (approved):
LGTM — Workflows are now showing up in the manifest 😅 

I don't want to hold up this PR any more but I think it'd be good to add some tests for auto-registering functions and types used by workflows

PR #36, opened 2022-05-28
Remove callback_id from the type we export to the manifest

shapirone:
# Summary
Make sure `callback_id` doesn't make it to the manifest

## Description
Right now if we try to use a type and deploy it, we'll get an error from the API endpoint that `callback_id` isn't a valid field for types. We need to make sure that `callback_id` is dropped before we export the manifest, which we now do by using an `export` method on the Custom Type.


selfcontained (approved):
Good find.

PR #38, opened 2022-06-03
Adding a workflows doc

selfcontained:
## Summary
Adding a `workflows.md` doc file that explains the basics of how a workflow can be defined. At this point, it's missing a callout to how you would setup a trigger to call a workflow, but that's pending that functionality landing still. In the vein of having a start to the workflows docs, and something we could point those that are interested in just defining workflows in code, I wanted to put this together.


shapirone (approved):
Excellent!

PR #40, opened 2022-06-17
Add type support for integer, number, usergroup, timestamp, and blocks

shapirone:
### Summary

Add type support for `integer`, `number`, `usergroup_id`, `timestamp`, and `blocks`

#### Details
Reintroducing support for types that aren't supported by input fields.

#### Testing
I'd recommend doing this somewhere with auto-generated shortcuts still enabled for every function, and then using these various types.


filmaj (approved):
Looks good and works great in my prod workspace! :shipit:

PR #41, opened 2022-06-23
added a Pending state to the function output type

jmplexi:
Added a PendingFunctionReturn type to allow functions to return the {completed: false} state. Unsure whether we should be allowing output along with completed: false in the future

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -111,28 +111,30 @@ export type BaseSlackFunctionHandler<
   | AsyncFunctionHandler<InputParameters, OutputParameters>
   | SyncFunctionHandler<InputParameters, OutputParameters>;
 
-type SuccessfulFunctionReturnArgs<
-  OutputParameters extends FunctionParameters,
-> = {
-  completed?: boolean;
-  // Allow function to return an empty object if no outputs are defined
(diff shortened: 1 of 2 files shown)


src/functions/types.ts
```diff
@@ -131,11 +131,18 @@ type ErroredFunctionReturnArgs<OutputParameters> =
   & Partial<SuccessfulFunctionReturnArgs<OutputParameters>>
   & Required<Pick<SuccessfulFunctionReturnArgs<OutputParameters>, "error">>;
 
+type PendingFunctionReturnArgs<OutputParameters> =
+  & {
+    completed: false;
+  }
+  & Partial<SuccessfulFunctionReturnArgs<OutputParameters>>;
```
shapirone, line 138: @selfcontained do we want this partial here or should we just allow `{ completed: false }`?

selfcontained: Just the completed: false - If a functions is return that, it indicates they’ll be calling an api to complete the function at a later time, and provide their outputs then. Providing outputs when returning completed: false wouldn’t have any effect with the outputs, so I don’t think we should signal they can return outputs too in that situation.

shapirone (changes requested):
Per Brad, let's remove that partial

PR #42, opened 2022-06-25
Adding support for the Object type again

shapirone:
### Summary
Reintroducing support for the `object` type

#### Details
This is built on top of the general `neil-readd-types` branch. It reintroduces support for:
- The type in `SchemaTypes`
- Using `object` as a parameter
- Object property workflow step output access
- Object property runtime access
- Unsupported objects (no defined `properties`)
- Manifest auto-registration for `types` and `datastores`

While verifying I identified two pre-existing bugs that I entered tickets for:
- [Deno SDK — Typeahead does not work for an Object parameter's `required` field](https://jira.tinyspeck.com/browse/HERMES-3054)
- [Deno Client — Datastore attributes of type object do not give typeahead](https://jira.tinyspeck.com/browse/HERMES-3055)

### DIFF
--- a/src/workflows/workflows_test.ts
+++ b/src/workflows/workflows_test.ts
@@ -15,6 +15,13 @@ Deno.test("WorkflowStep export input values", () => {
         name: {
           type: "string",
         },
+        manager: {
+          type: "object",
+          properties: {
+            email: { type: "string" },
+            name: { type: "string" },
+          },
+        },
       },
       required: ["email"],
     },
@@ -39,8 +46,15 @@ Deno.test("WorkflowStep export input values", () => {
         name: {
           type: "string",
         },
+        manager: {
+          type: "object",
+          properties: {
+            email: { type: "string" },
+            name: { type: "string" },
+          },
+        },
       },
-      required: ["email"],
+      required: ["email", "manager"],
     },
   });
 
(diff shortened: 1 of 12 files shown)


src/workflows/workflows_test.ts
```diff
@@ -74,6 +99,8 @@ Deno.test("WorkflowStep export input values", () => {
   assertEquals(`${step1Inputs.email}`, "{{inputs.email}}");
   assertEquals(`${step1Inputs.name}`, "A name: {{inputs.name}}");
   assertEquals(`${step1.outputs.url}`, "{{steps.0.url}}");
+  assertEquals(`${step1.outputs.manager?.email}`, "{{steps.0.manager.email}}");
```
selfcontained, line 102: I noticed this when testing that the output parameter reference is treated as optional here, and shouldn't be (these will always be set for parameter references, it's runtime where it's optional). I'll open up an issue for what needs fixed here, I think it's a slight tweak to the `WorkflowParameterReferences` type.

selfcontained (approved):
Nice job! Works as expected. One thing I noticed was for runtime input typing, for an object w/ a few properties, if you use a custom type reference (result of `DefineType`) vs. an inline object type on the function input parameter, at runtime the parameter is an untyped object when using the DefineType reference, but it's fully typed for the inline object definition. I don't think this is blocking, but wanted to mention it.

src/workflows/workflows_test.ts
selfcontained: Issue created - #48

shapirone:
> Nice job! Works as expected. One thing I noticed was for runtime input typing, for an object w/ a few properties, if you use a custom type reference (result of `DefineType`) vs. an inline object type on the function input parameter, at runtime the parameter is an untyped object when using the DefineType reference, but it's fully typed for the inline object definition. I don't think this is blocking, but wanted to mention it.

https://github.com/slackapi/deno-slack-sdk/issues/49 entered for this

PR #46, opened 2022-06-28
added tests to the functionhandler

jmplexi:
###  Summary

Added tests to the functionhandler to account for completed: false output and error only output

### DIFF
--- a/src/functions/base_function_handler_type_test.ts
+++ b/src/functions/base_function_handler_type_test.ts
@@ -103,6 +103,34 @@ Deno.test("BaseSlackFunctionHandler with any inputs and any outputs", () => {
   assertEquals(result.outputs?.out, inputs.in);
 });
 
+Deno.test("BaseSlackFunctionHandler with no inputs and error output", () => {
+  // deno-lint-ignore no-explicit-any
+  const handler: BaseSlackFunctionHandler<any, any> = () => {
+    return {
+      error: "error"
+    };
+  };
+  const { createContext } = SlackFunctionTester("test");
+  const inputs = { in: "test" };
+  const result = handler(createContext({ inputs }));
+  assertEquals(result.error, "error");
+});
+
+Deno.test("BaseSlackFunctionHandler with no inputs and completed false output", () => {
+  // deno-lint-ignore no-explicit-any
+  const handler: BaseSlackFunctionHandler<any, any> = () => {
+    return {
+      completed: false,
+    };
+  };
+  const { createContext } = SlackFunctionTester("test");
+  const inputs = { in: "test" };
+  const result = handler(createContext({ inputs }));
+  assertEquals(result.completed, false);
+});
+
+
+
 Deno.test("BaseSlackFunctionHandler with set inputs and any outputs", () => {
   type Inputs = {
     in: string;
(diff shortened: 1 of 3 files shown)


src/functions/slack_function_handler_type_test.ts
```diff
@@ -190,3 +190,51 @@ Deno.test("SlackFunctionHandler with only outputs", () => {
...
+  const TestFn = DefineFunction({
+    callback_id: "test",
+    title: "test fn",
+    source_file: "test.ts",
+    output_parameters: {
+      properties: {
+        completed: {
+          type: "boolean",
+        },
+      },
```
shapirone, line 204: nit: Naming the output param `completed` is a bit confusing. I think it's a good idea for this test to have output parameters, and actually would be even better if those outputs are required. Let's just name it something else and include it in the `required` array.

The reason for this is that a Successful response would look like
```ts
return {
  outputs: {
    completed: false,
  }
}
```

But what we want to test for is what you have already
```ts
return {
  completed: false,
}
```

IMO these are too similar and might trip someone up.

src/functions/slack_function_handler_type_test.ts
```diff
@@ -190,3 +190,51 @@ Deno.test("SlackFunctionHandler with only outputs", () => {
...
+    title: "test fn",
+    source_file: "test.ts",
+    output_parameters: {
+      properties: {
+        error: {
+          type: "string",
+        },
+      },
+      required: [],
+    },
```
shapirone, line 230: nit: Same comment as above where we don't want the function to have a valid output named the same thing we're testing for.

src/functions/base_function_handler_type_test.ts
```diff
@@ -103,6 +103,32 @@ Deno.test("BaseSlackFunctionHandler with any inputs and any outputs", () => {
...
+Deno.test("BaseSlackFunctionHandler with no inputs and error output", () => {
+  // deno-lint-ignore no-explicit-any
+  const handler: BaseSlackFunctionHandler<any, any> = () => {
+    return {
+      error: "error",
+    };
+  };
+  const { createContext } = SlackFunctionTester("test");
+  const inputs = { in: "test" };
+  const result = handler(createContext({ inputs }));
```
shapirone, line 115: The test title says that this doesn't have any `inputs`, so we can set it to an empty object here like `const result = handler(createContext({ inputs: {} }));`

src/functions/base_function_handler_type_test.ts
```diff
@@ -103,6 +103,32 @@ Deno.test("BaseSlackFunctionHandler with any inputs and any outputs", () => {
...
+Deno.test("BaseSlackFunctionHandler with no inputs and completed false output", () => {
+  // deno-lint-ignore no-explicit-any
+  const handler: BaseSlackFunctionHandler<any, any> = () => {
+    return {
+      completed: false,
+    };
+  };
+  const { createContext } = SlackFunctionTester("test");
+  const inputs = { in: "test" };
+  const result = handler(createContext({ inputs }));
```
shapirone, line 128: `inputs` should also be empty here based on the test title

shapirone (changes requested):

shapirone (approved):
LGTM 🎉

PR #47, opened 2022-07-01
Add Non-Slack hosted app manifest types

neptunel:
## Summary

The goal of this PR is to expand the `SlackManifest` type definition to include all currently accepted manifest fields. 

Broadly, changes are 
* Added new `src/manifest` folder and relocated SlackManifest class, types, and ManifestSchema and tests here
   * `/manifest_schema.ts`: Contains underlying ManifestSchema types which directly map to internal types. Prefixed with `Manifest`.
   * `/types.ts`: Contains SlackManifestType and new Hosting vs. Remote specific interfaces. 
       * Now there's a new `runOnSlack` property. Remote Slack app developers will explicitly set this as false.

* Updates on files outside of `src/manifest/*` are largely updating imports!

### Updates

- `SlackManifestType` changed to being a discriminated union type of `ISlackManifestHosted` and `ISlackManifestRemote` interfaces, where the discriminant property `slackHosted` maps to `function_runtime` in the underlying `ManifestSchema ` to contain both remote & slack hosted app features. 

-  To provide a better developer experience, instead of relying on API validation to let developers know they got it wrong when a dev trying to add an unsupported type, the `slackHosted` type was added to require the developer to be explicit when they're defining the manifest that it's slack hosted/remote. `slackHosted`= true by default and maps to `function_runtime` = "slack" in `ManifestSchema`. `slackHosted`= false maps to function_runtime = "remote" in ManifestSchema. 

- Previous types in deno-slack-sdk still exist under `ISlackManifestHosted`, in addition to these types remote app features were added under `ISlackManifestRemote`. 

- manifest_schema.ts: `ManifestSchema` updated with newly added features such as `function_runtime`, `settings`, `app_directory`; and `app_home`, `shortcuts`, `slash_commands`, `unfurl_domains`, `workflow_steps` under `features`. Also `oauth_config` now supports `user` scopes, `redirect_urls` and `token_management_enabled` in addition to the bot scopes. 

- The `short_description` under `display_information` was changed to  `description` as it was named description in the webapp app-manifests.

### Considerations 

- Omitted `functions` and `schema` props from the `ManifestSchema.features` because they are redundant with top-level schema properties.
- Omitted `ManifestSchema.forms` because this appears to be a feature under development
- Omitted `ManifestSchema.compliance`. This is GovSlack-only property.
- Omitted `ManifestSchema.manifest_metadata` property (can be added later) 
- Omitted `hermes_app_type` because this is no longer used (`function_runtime` is used instead)

## Testing

1. Build deno-slack-sdk as a node module

- Pull down [deno-slack-sdk#expand-manifest-types-fekiz](https://github.com/slackapi/deno-slack-sdk/tree/expand-manifest-types-fekiz) locally

- Run `deno run -A build_npm.ts 0.1.0 && cd npm && npm link`, to build the npm distribution folder and generate a symlink

2. Link this @slack/deno-slack-sdk module in a test Bolt JS project

- Pull down [bolt-js-starter-template#future-fekiz-test](https://github.com/slack-samples/bolt-js-starter-template/tree/future-fekiz-test)

- Install dependencies and symlink:`npm i && npm link @slack/deno-slack-sdk`

3. Test the updated manifest

- Currently `hermes manifest` will not show the manifest.json properly because some of the types and fields are missing in the [CLI app_manifest.go](https://github.com/slackapi/slack-cli/blob/6bb3de4b9226ec33842a05c0872c322ee98e3035/internal/shared/types/app_manifest.go#L51) and it omits some fields with a `false` value. (Ex: the `ommitempty` on [this line](https://github.com/slackapi/slack-cli/blob/6bb3de4b9226ec33842a05c0872c322ee98e3035/internal/shared/types/app_manifest.go#L79) is why `false` for `always_online` disappears in the manifest.json ). We need to update the missing fields in the CLI app_manifest.go according to the types in the deno-slack-sdk-#expand-manifest-types-fekiz, to be able to use `hermes manifest`.

- So, instead of using `hermes manifest` please try to use `./node_modules/@slack/bolt/src/cli/get-manifest.js` after successfully linking the deno-slack-sdk to bolt-js-starter-template.

* [ x ] I've read and understood the [Contributing Guidelines](https://github.com/{project_slug}/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
* [ x ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).


src/manifest.ts
```diff
@@ -23,11 +23,13 @@ export class SlackManifest {
...
         "major_version": 2,
       },
+
+      "function_runtime": this.getFunctionRuntime(),
       "display_information": {
         background_color: def.backgroundColor,
         name: def.name,
         long_description: def.longDescription,
-        short_description: def.description,
+        description: def.description,
```
stevengill, line 32: This was previously broken

src/manifest.ts
```diff
@@ -23,11 +23,13 @@ export class SlackManifest {
         // todo: is there a more idiomatic way of defining this? constant file?
         "major_version": 2,
       },
+
+      "function_runtime": this.getFunctionRuntime(),
```
stevengill, line 27: This currently introduces manifest validation issues as support for this isn't in the backend yet correct?

src/manifest.ts
```diff
@@ -41,9 +43,67 @@ export class SlackManifest {
...
+        manifest.features.slash_commands = def.features?.slashCommands;
+      }
+      if (def.features?.unfurlDomains !== undefined) {
+        manifest.features.unfurl_domains = def.features?.unfurlDomains;
+      }
+      if (def.features?.workflowSteps !== undefined) {
+        manifest.features.workflow_steps = def.features?.workflowSteps;
+      }
+    } else {
+      manifest.outgoing_domains = def.outgoingDomains || [];
```
stevengill, line 104: Nice! outgoing domains are only relevant if app is hosted on ROSI

src/manifest.ts
```diff
@@ -41,9 +43,67 @@ export class SlackManifest {
             def.name,
         },
       },
-      "outgoing_domains": def.outgoingDomains || [],
+      settings: {},
     } as ManifestSchema;
 
+    if (def.slackHosted == false) {
```
stevengill, line 49: what happens when `def.slackHosted` === undefined?

stevengill: maybe instead of using `def.slackHosted` here, we should use `manifest.function_runtime` as it handles the undefined usecase (sets it to true by default)

src/types.ts
```diff
@@ -6,17 +6,32 @@ import type {
   ParameterSetDefinition,
 } from "./parameters/mod.ts";
 import type { ICustomType } from "./types/types.ts";
-
-// SlackManifestType is the top level type that imports all resources for the app
-// An app manifest is generated based on what this has defined in it
+// import type { CamelCasedPropertiesDeep } from "./types_utils/camel-case.ts";
```
stevengill, line 9: Why did you have to comment out this import again?

src/types_utils/camel-case.ts
```diff
@@ -0,0 +1,47 @@
+/** Utility types to enable conversion of type properties to camelCase
```
stevengill, line 1: Does this file still need to be in here?

src/types.ts
```diff
@@ -29,7 +44,70 @@ export type SlackManifestType = {
...
+  userScopes?: Array<string>;
+  redirectUrls?: Array<string>;
+  tokenManagementEnabled?: boolean;
+
+  features?: ISlackManifestRemoteFeaturesSchema;
+}
+
+export interface ISlackManifestRemoteFeaturesSchema {
+  appHome?: ManifestAppHomeSchema;
+  botUser?: Omit<ManifestBotUserSchema, "display_name">;
```
stevengill, line 96: Omitting `dispaly_name` here since it is available at the top level

src/types.ts
```diff
@@ -29,7 +44,70 @@ export type SlackManifestType = {
...
+  slashCommands?: ManifestSlashCommandsSchema;
+  unfurlDomains?: ManifestUnfurlDomainsSchema;
+  workflowSteps?: ManifestWorkflowStepsSchema;
+}
+
+/** SlackManifestFromType is a type function which takes a parameterT.
+ * It looks at the SlackManifestType which is a discriminated union.
+ * The function returns the type whose discriminant property (slackHosted) matches T
+ */
+export type SlackManifestFromType<T extends boolean> = Extract<
```
stevengill, line 107: This isn't used anywhere from what I can see. But it could be useful for devs

src/manifest.ts
neptunel: Yes, currently` function_runtime` is under settings schema in the [ `webapp app_manifest`](https://slack-github.com/slack/webapp/blob/master/js/modern/generated/types/api/schemas/app-manifests/v1/settings/schema.ts). To be able to validate ` function_runtime` in our `ManifestSchema`, we also have to update the[ `webapp manifest.ts`](https://slack-github.com/slack/webapp/blob/master/js/modern/generated/types/api/schemas/app-manifests/v1/manifest.ts)

src/manifest.ts
neptunel: Yes, this is called `description` under the [`webapp display_information schema`](https://slack-github.com/slack/webapp/blob/master/js/modern/generated/types/api/schemas/app-manifests/v1/display_information/schema.ts)`, so changed the ``short_description`` to  description to prevent invalidation.

src/types_utils/camel-case.ts
neptunel: I think we can keep this file until we change its behavior so that it won't hurt `SlackManifestType`  discriminated union or decide how we going to handle the case conversions (snake to camel case).

src/manifest.ts
stevengill: Should we move `function_runtime` under settings here instead? (Keep it top level but move it under settings when generating the manifest)

src/types.ts
neptunel: I don't have to actually, it's just not in the use yet.

src/manifest.ts
neptunel: Using `manifest.function_runtime`  is breaking the union type to be ` ISlackManifestRemote` under its current scope. So to handle` def.slackHosted === undefined` issue I changed it to `(def.slackHosted === false && def.slackHosted !== undefined)`

src/types.ts
neptunel: Maybe we may need to use this function when we trying to change the camel case file in the future, to keep the `SlackManifestType` act according to the discriminant `slackHosted`.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,46 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Build & Deploy to NPM
+
+on:
+  push:    
+    tags:        
+      - '*'
+    branches: [ main ]
```
srajiang, line 10: I see that this is set up to run every time there is a push to `main`, which is fine for testing purposes. 

However, in production means that these modules will be published to npm possibly _before_ they are published to deno.land, since commits `main` are not synced with the release schedule currently. 

Instead I recommend that we trigger this Github workflow every time that a [release](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) event occurs. When Neil or Brad are updating this module, they will publish a release. You can see a list of previous releases [here](https://github.com/slackapi/deno-slack-sdk/releases)

You can also see from the event payload that that a GITHUB_REF value is provided, this you can use in line 33 instead of a hardcoded value (0.1.0) so that the version we publish to NPM also matches the version published to deno.land.

build_npm.ts
```diff
@@ -0,0 +1,31 @@
...
+  outDir: "./npm",
+  shims: {
+    // see JS docs for overview and more options
+    deno: true,
+  },
+  package: {
+    // package.json properties
+    name: "@slack/deno-slack-sdk",
+    version: Deno.args[0],
+    description: "Test deno sdk dpendencies",
```
srajiang, line 17: Let's modify this general package description to be more descriptive for external users. Once the package is published to npm, developers will see as the description for this package,.

srajiang (changes requested):
Hi Sena! Breaking up review into a couple of parts. In this first pass, my comments are around the npm-publish.yml file. and the setup of the package. 

Can you also include some testing instructions for how to test this github action workflow?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: macos-latest
+    
+    steps:
+      - name: Actions checkout
+        uses: actions/checkout@v3
+      
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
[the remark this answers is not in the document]
neptunel: The latest would work, version: 16 was in the templates, I'll update this.

srajiang (changes requested):
Second pass of comments with some suggestions for improvements. I haven't done any testing yet on this.

src/types.ts
```diff
@@ -6,36 +6,35 @@ import type {
...
-};
+/** ISlackManifestHosted contains the features currently available to
+ * ManifestSchema#function_runtime = slack hosted apps. Here, we surface that value
+ * to the developer as the slackHosted top-level property.
+ */
 
-export interface SlackManifestFeatures {
-  appHome?: SlackManifestFeaturesAppHome;
-}
+//  features?: SlackManifestFeatures;
```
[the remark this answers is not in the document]
neptunel: Yup, removed it!

shapirone (changes requested):
Awesome! I took a quick look and added some comments throughout the review. A couple high level points:
- I'm having a hard time what type goes to which schema and I think it's because the `src/types.ts` file is becoming really big and juggling multiple concepts. I think it'd be great to create new files to separate it out. Something like a new `manifest` folder with `src/manifest/mod.ts` replacing `src/manifest.ts` and for the types we can have `src/manifest/types.ts` handle the shared types while`src/manifest/remoteTypes.ts`, `src/manifest/hostedTypes.ts` are for their own unique types. We can still export them from the top level `src/types.ts` if we want to surface them to developers more directly!
- Let's standardize on removing string wrapping in keys i.e. `name: string` instead of `"name": string;`

src/types.ts
```diff
@@ -141,43 +205,181 @@ export type ManifestWorkflowSchema = {
...
+  name: string;
+  type: "message" | "global";
+  "callback_id": string;
+  description: string;
+};
+
+export type ManifestShortcutsSchema = [
+  ManifestShortcutSchema,
+  ...ManifestShortcutSchema[],
+];
```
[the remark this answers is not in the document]
neptunel: Yes, but this notation is specifically for the array requiring at least 1 `ManifestShortcutSchema` type in it.  This is also how it's used in the [`webapp`](https://slack-github.com/slack/webapp/blob/master/js/modern/generated/types/api/schemas/app-manifests/v1/features/shortcuts.ts)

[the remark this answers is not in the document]
shapirone: Ohh, that makes perfect sense! Since we do this a lot, maybe a new Utility Type that indicates what it's doing would help? I.e.
```ts
type PopulatedArray<T> = [T, ...T[]];

export type ManifestShortcutsSchema = PopulatedArray<ManifestShortcutSchema>;
```

src/types.ts
```diff
@@ -91,20 +153,22 @@ export type Env = Record<string, string>;
...
-  input_parameters: ManifestFunctionParameters;
-  output_parameters: ManifestFunctionParameters;
+  "source_file": string;
+  "input_parameters": ManifestFunctionParameters;
+  "output_parameters": ManifestFunctionParameters;
 };
 
+export type ManifestCustomTypeSchema = ParameterDefinition;
+
+// TODO: Brand new type below
```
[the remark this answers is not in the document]
neptunel: Removing it!

src/manifest_test.ts
```diff
@@ -352,15 +354,225 @@ Deno.test("SlackManifest.export() will not duplicate datastore scopes if they're
...
+    definition.features?.botUser?.always_online,
+  );
+  assertStrictEquals(
+    manifest.features.shortcuts,
+    definition.features?.shortcuts,
+  );
+  assertEquals(manifest.settings.function_runtime, "remote");
+});
+
+// Deno.test("Manifest() property mappings fo expanded types in the remote manifest", () => {
```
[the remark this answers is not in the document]
neptunel: Yes, I'll be using them to test remote app types.

src/types.ts
```diff
@@ -91,20 +153,22 @@ export type Env = Record<string, string>;
...
 export type ManifestFunctionSchema = {
+  //title and description is optional ->source file does not exists
   title?: string;
   description?: string;
-  source_file: string;
-  input_parameters: ManifestFunctionParameters;
-  output_parameters: ManifestFunctionParameters;
+  "source_file": string;
+  "input_parameters": ManifestFunctionParameters;
+  "output_parameters": ManifestFunctionParameters;
```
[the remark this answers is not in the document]
neptunel: Removed all double quotes

src/types.ts
```diff
@@ -91,20 +153,22 @@ export type Env = Record<string, string>;
...
   [index: number]: string | number | symbol;
 };
-
 export type ManifestFunctionParameters = {
   required?: RequiredParameters;
   properties: ParameterSetDefinition;
 };
-
 export type ManifestFunctionSchema = {
+  //title and description is optional ->source file does not exists
```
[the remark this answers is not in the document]
neptunel: That's commented before I started to work on it by @srajiang.  But, I think it meant if `source_file`(
required field) does not exist, `title` and `description` should not be required, so they are set as optional.

src/manifest.ts
```diff
@@ -48,8 +48,67 @@ export class SlackManifest {
           messages_tab_read_only_enabled: true,
         },
       },
-      outgoing_domains: def.outgoingDomains || [],
-    };
+      settings: { "function_runtime": this.getFunctionRuntime() },
+    } as ManifestSchema;
+
+    if (def.slackHosted === false && def.slackHosted !== undefined) {
+      //Settings
```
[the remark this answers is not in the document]
neptunel: Done! Thanks!

src/types.ts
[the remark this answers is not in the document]
neptunel: Removed it since it was confusing!

src/manifest.ts
```diff
@@ -48,8 +48,67 @@ export class SlackManifest {
...
+      if (def.userScopes !== undefined) {
+        manifest.oauth_config.scopes.user = def.userScopes;
+      }
+      if (def.redirectUrls !== undefined) {
+        manifest.oauth_config.redirect_urls = def.redirectUrls;
+      }
+      if (def.tokenManagementEnabled !== undefined) {
+        manifest.oauth_config.token_management_enabled =
+          def.tokenManagementEnabled;
+      }
```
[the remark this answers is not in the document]
neptunel: Removed the undefined checks for these types in my latest commit in my fork

src/manifest.ts
```diff
@@ -48,8 +48,67 @@ export class SlackManifest {
...
+      //Features
+      if (def.features?.appHome !== undefined) {
+        manifest.features.app_home = def.features?.appHome;
+      }
```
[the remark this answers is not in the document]
neptunel: It created conflicts since the `ManifestFeaturesSchema` in ManifestSchema was still including the old app home schema. Updated it with `ManifestAppHomeSchema`, so the logic added on the last release to AppHome became common in both slack-hosted and remote types. Also, existing logic in [this file ](https://github.com/slackapi/deno-slack-sdk/blob/expand-manifest-types-fekiz/src/manifest.ts#L153) updated with `home-tab_enabled` addition to override any existing `app_home` defaults if supplied in the manifest. And, failing test cases due to naming convention inconsistencies such as using camelCase with defining lower-level properties fixed. Camel case will only be in use at the upper level `ISlackManifestRemote` & `ISlackManifestHosted` types. Other Schema types as well as ManifestSchema are named using snake case.

src/types.ts
```diff
@@ -6,36 +6,35 @@ import type {
...
+ * to the developer as the slackHosted top-level property.
+ */
 
-export interface SlackManifestFeatures {
-  appHome?: SlackManifestFeaturesAppHome;
-}
+//  features?: SlackManifestFeatures;
 
-export type SlackManifestType = {
+export interface ISlackManifestHosted {
```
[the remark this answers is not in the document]
srajiang: Great suggestion, it'll also make it more clear as changes / adjustments are made later on down the line. 
I've just done a pass doing some tidying up structurally and included this in latest changes.

srajiang (commented):
Great job @neptunel with getting the bulk of these new types in and for staying on top of the myriad features in platform in the process! This was no small effort, especially in terms of reconciling existing typing. 

I've pushed changes per @shapirone's suggestions around high level organization which I think will make this set of files more maintainable in the long run. 

These include:
* Reorganizing what used to be in `src/manifest.ts` into a separate directory
* Relocating manifest related types from `src/types.ts`.
* Creating an ISlackManifestShared to contain types and features that are common to both SlackHosted and Remote app scenarios
* Relocated all ManifestSchema types into its own file. Since these pertain directly to internal types it made sense to keep them in their own little corner. 


Thanks reviewers for your attention to this PR so far. Would love a second round of 👀 at this point.

shapirone (changes requested):
This is awesome! I left a few nitpick comments, but my big concern is the snake case vs camel case problem. Generally for the Deno SDK we have taken in camel case and output snake case. I think there's still some structural improvements we could do to separate out inputs to our `Manifest` function (which should be camel case) and what we expect that function to output (which should be snake case), but I don't think that needs to be done in this PR. @srajiang I'd need your input on whether or not that's problematic for Bolt JS devs (if they are used to snake case)

src/manifest/types.ts
```diff
@@ -0,0 +1,95 @@
...
+  | ISlackManifestRemote;
+
+/** Slack-hosted app manifest
+ *
+ * When runOnSlack = true
+ * Corresponds to function_runtime = slack in ManifestSchema.
+ */
+export interface ISlackManifestRunOnSlack extends ISlackManifestShared {
+  runOnSlack?: true; // maps to function_runtime = "slack" in ManifestSchema, optional since the apps are slack hosted by default
+  outgoingDomains?: Array<string>;
```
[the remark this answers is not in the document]
srajiang: Good catch - Removed!

src/manifest/mod.ts
```diff
@@ -93,29 +99,27 @@ export class SlackManifest {
...
-
-      if (messagesTabEnabled !== undefined) {
-        manifest.features.app_home.messages_tab_enabled = messagesTabEnabled;
-      }
-      if (messagesTabReadOnlyEnabled !== undefined) {
+      const {
+        home_tab_enabled,
+        messages_tab_enabled,
+        messages_tab_read_only_enabled,
+      } = def.features.appHome;
```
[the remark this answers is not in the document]
srajiang: @shapirone. Yeah good catch. These are snake case now because the underlying `ManifestAppHomeSchema` is in snake_case (in following with how ManifestSchema) is serialized. I think types in `manifest_schema.ts` types should match the output, so I'm not super inclined to change this back. 

Folks who do have this as camelCase currently would need to correct the type now as part of these changes if they have it in their manifest, though I don't think that should be too big of a problem for devs. Maybe we can add a release note? 

The bigger vision to have devs work in camelCase and conversion to snake_case happening under the hood I agree with completely though. To that end, I noticed camelCase vs. snake_case is used interchangeably in our reverseString sample, so it would be good to revisit that holistically in a separate PR. Depending on how deeply we decide to make casing consistent, it could raise type errors in existing apps (say in a DefineFunction),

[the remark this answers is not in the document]
srajiang: Added a `CamelCased` conversion to resolve this issue and synced about this offline!

src/manifest/mod.ts
```diff
@@ -185,4 +191,60 @@ export class SlackManifest {
...
+    //Settings
+    manifest.settings = {
+      ...manifest.settings,
+      ...def.settings,
+    };
+    manifest.settings.event_subscriptions = def.eventSubscriptions;
+    manifest.settings.socket_mode_enabled = def.socketModeEnabled;
+    manifest.settings.token_rotation_enabled = def.tokenRotationEnabled;
+
+    console.log(manifest.settings);
```
shapirone, line 214: 👁️👁️

srajiang: Good catch! Removed, thanks!

shapirone (approved):
Monumental! 🎉  Great work y'all, let's merge this after we do tomorrow's deploy 🙌

PR #50, opened 2022-07-06
Add new name property to DefineType definition

jmplexi:
###  Summary

Add new name property to DefineType definition

### DIFF
--- a/src/manifest_test.ts
+++ b/src/manifest_test.ts
@@ -99,6 +99,100 @@ Deno.test("Manifest() automatically registers types used by function input and o
   });
 });
 
+Deno.test("Manifest() properly converts name to proper key", () => {
+  const UsingName = DefineType({
+    name: "Using Name",
+    type: Schema.types.boolean,
+  });
+
+  const Function = DefineFunction(
+    {
+      callback_id: "test_function",
+      title: "Function title",
+      source_file: "functions/test_function.ts",
+      input_parameters: {
+        properties: { aType: { type: UsingName } },
+        required: [],
+      },
+    },
+  );
+
+  const definition: SlackManifestType = {
+    name: "Name",
+    description: "Description",
+    icon: "icon.png",
+    longDescription: "LongDescription",
+    botScopes: [],
+    functions: [Function],
+  };
+  const manifest = Manifest(definition);
+  assertEquals(manifest.types, { "Using Name": { type: "boolean" } });
+});
+
+Deno.test("Manifest() properly converts callback_id to proper key", () => {
+  const UsingCallback = DefineType({
+    callback_id: "Using Callback",
+    type: Schema.types.boolean,
+  });
+
+  const Function = DefineFunction(
+    {
+      callback_id: "test_function",
+      title: "Function title",
+      source_file: "functions/test_function.ts",
+      input_parameters: {
+        properties: { aType: { type: UsingCallback } },
+        required: [],
+      },
+    },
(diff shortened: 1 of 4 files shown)


src/types/types.ts
```diff
@@ -3,8 +3,14 @@ import { SlackManifest } from "../manifest.ts";
 import { ManifestCustomTypeSchema } from "../types.ts";
 
 export type CustomTypeDefinition =
-  & { callback_id: string }
-  & TypedParameterDefinition;
+  | (
+    & { callback_id: string }
```
shapirone, line 7: Is it possible to mark `callback_id` deprecated like we do [here](https://github.com/slackapi/deno-slack-sdk/blob/main/src/functions/types.ts#L131)?

src/types/mod.ts
```diff
@@ -53,8 +58,15 @@ export class CustomType<Def extends CustomTypeDefinition>
...
+    // remove callback_id or name from the definition we pass to the manifest
+    if ("callback_id" in this.definition) {
+      const { callback_id: _c, ...definition } = this.definition;
+      return definition;
+    } else if ("name" in this.definition) {
+      const { name: _n, ...definition } = this.definition;
+      return definition;
+    } else {
+      return this.definition;
+    }
```
shapirone, line 70: Similar to above, we can rely on there only being two options.

```ts
if ("callback_id" in this.definition) {
      const { callback_id: _c, ...definition } = this.definition;
      return definition;
    } else {
      const { name: _n, ...definition } = this.definition;
      return definition;
    }
```

src/types/types_test.ts
```diff
@@ -0,0 +1,84 @@
...
+    title: "Title",
+    description: "Description",
+    callback_id: "Callback_id",
+    type: "Type",
+  });
+
+  assertEquals(Type.id, "Callback_id");
+});
+
+Deno.test("DefineType test topString using the callback_id parameter", () => {
```
shapirone, line 26: typo: `toString` rather than `topString`

shapirone (changes requested):
Awesome, this is looking great! I think we should simplify a few things here and see if we can mark `callback_id` deprecated.

Also, it'd be great to add a test to the `manifest_test.ts` file to ensure that both `callback_id` and `name` are converted to the key appropriately in the Manifest fn response.

src/manifest_test.ts
```diff
@@ -99,6 +99,100 @@ Deno.test("Manifest() automatically registers types used by function input and o
...
+    {
+      callback_id: "test_function",
+      title: "Function title",
+      source_file: "functions/test_function.ts",
+      input_parameters: {
+        properties: { aType: { type: UsingName } },
+        required: [],
+      },
+    },
+  );
```
shapirone, line 118: We don't need this function. Instead we can add `types: [UsingName]` to the manifest to keep the scope of the test limited

src/types/types.ts
```diff
@@ -3,8 +3,19 @@ import { SlackManifest } from "../manifest.ts";
 import { ManifestCustomTypeSchema } from "../types.ts";
 
 export type CustomTypeDefinition =
-  & { callback_id: string }
-  & TypedParameterDefinition;
+  | (
+    & {
+      /**
+       * @deprecated Use name instead
+       */
```
shapirone, line 10: This isn't doing what I would expect as it's not messaging the user as they use `DefineType`. We can come back to it in another PR.

PR #55, opened 2022-07-08
Support for read_only_message_tab

abhishek199-dhn:
###  Summary

Adding a support to enable message tab in the app profile.

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -1,7 +1,11 @@
 import { ParameterSetDefinition } from "./parameters/mod.ts";
 import {
+  ManifestCustomTypesSchema,
+  ManifestDataStoresSchema,
   ManifestFunction,
+  ManifestFunctionsSchema,
   ManifestSchema,
+  ManifestWorkflowsSchema,
   SlackManifestType,
 } from "./types.ts";
 import { ICustomType } from "./types/types.ts";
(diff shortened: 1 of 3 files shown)


src/manifest.ts
```diff
@@ -18,58 +22,77 @@ export class SlackManifest {
...
+          acc[datastore.name] = datastore.export();
+          return acc;
+        },
+        {},
+      );
+    }
+
+    if (def.features?.appHome) {
+      manifest.features.app_home = {
+        messages_tab_enabled: def.features?.appHome.messagesTabEnabled,
```
selfcontained, line 92: Should we be providing a default value here as well, or is there a reason we want to intentionally not default this?

src/manifest.ts
```diff
@@ -18,58 +22,77 @@ export class SlackManifest {
...
+      manifest.datastores = def.datastores?.reduce<ManifestDataStoresSchema>(
+        (acc = {}, datastore) => {
+          acc[datastore.name] = datastore.export();
+          return acc;
+        },
+        {},
+      );
+    }
+
+    if (def.features?.appHome) {
```
shapirone, line 90: This means that `app_home` won't be set unless the developer specifies something at `features.appHome`

shapirone (changes requested):

src/manifest.ts
shapirone: It seems like we need to default this to `true` or else the Messages tab won't appear at all.

shapirone (approved):
Thanks @abhishek199-dhn!

PR #57, opened 2022-07-11
First pass at Action Handlers / a router for actions

filmaj:
###  Summary

Mostly following the template experimented with here: https://corp.quip.com/FF6hAO3C4UbL/Function-Interactivity-Examples

Pair this up with the PR on internal GitHub for the app consuming this: https://slack-github.com/bharris/interactive-approval/pull/1

Also loosely based on[ @selfcontained's `bradh-fn-interactivity-handlers-duex` branch of deno-slack-runtime](https://github.com/slackapi/deno-slack-runtime/tree/bradh-fn-interactivity-handlers-duex).

* [ ] I've read and understood the [Contributing Guidelines](https://github.com/{project_slug}/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
* [ ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/src/functions/routers/action_router.ts
+++ b/src/functions/routers/action_router.ts
@@ -133,8 +133,10 @@ export class ActionsRouter<
         // Assumes an object as a constraint (type BlockActionConstraintObject)
         // Return first match *within* any of the defined fields on the constaint object, but ensure there is a match on *all* defined fields
         // Effectively a logical AND across the action_id and block_id field(s)
-        let actionIDMatched = false;
-        let blockIDMatched = false;
+        // If either of the constraint fields are not defined, pre-set them to have matched so we can effectively
+        // ignore them when determining if we have a match by &&'ing them
+        let actionIDMatched = constraint.action_id ? false : true;
+        let blockIDMatched = constraint.block_id ? false : true;
         if (constraint.action_id) {
           actionIDMatched = matchBlockActionConstraintField(
             normalizeConstraintToArray(constraint.action_id),
@@ -149,12 +151,7 @@ export class ActionsRouter<
             action,
           );
         }
-        if (
-          (constraint.block_id && constraint.action_id && blockIDMatched &&
-            actionIDMatched) ||
-          (constraint.block_id && blockIDMatched) ||
-          (constraint.action_id && actionIDMatched)
-        ) {
+        if (blockIDMatched && actionIDMatched) {
           return handler;
         }
       }
--- a/src/functions/routers/action_router.ts
+++ b/src/functions/routers/action_router.ts
@@ -1,0 +1,191 @@
+import {
+  ParameterSetDefinition,
+  PossibleParameterKeys,
+} from "../../parameters/mod.ts";
+import { SlackFunction } from "../mod.ts";
+import type {
+  BlockAction,
+  BlockActionConstraint,
+  BlockActionConstraintField,
+  BlockActionConstraintObject,
+  BlockActionHandler,
+} from "./types.ts";
+
+/**
+ * Define an actions "router" and its input and output parameters for use in a Slack application. The ActionsRouter will route incoming action events to action-specific handlers.
+ * @param {SlackFunction<InputParameters, OutputParameters, RequiredInput, RequiredOutput>} func Reference to your previously-created SlackFunction, defined via DefineFunction
+ * @returns {ActionsRouter}
+ */
+export const BlockActionsRouter = <
+  InputParameters extends ParameterSetDefinition,
+  OutputParameters extends ParameterSetDefinition,
+  RequiredInput extends PossibleParameterKeys<InputParameters>,
+  RequiredOutput extends PossibleParameterKeys<OutputParameters>,
+>(
+  func: SlackFunction<
+    InputParameters,
+    OutputParameters,
+    RequiredInput,
+    RequiredOutput
+  >,
+) => {
+  const router = new ActionsRouter(func);
+
+  // deno-lint-ignore no-explicit-any
+  const exportedHandler: any = router.export();
+
+  // deno-lint-ignore no-explicit-any
+  exportedHandler.addHandler = ((...args: any) => {
+    router.addHandler.apply(router, args);
+
+    return exportedHandler;
+  }) as typeof router.addHandler;
+
+  return exportedHandler as
+    & ReturnType<typeof router.export>
+    & Pick<typeof router, "addHandler">;
+};
+
+export class ActionsRouter<
+  InputParameters extends ParameterSetDefinition,
+  OutputParameters extends ParameterSetDefinition,
+  RequiredInput extends PossibleParameterKeys<InputParameters>,
+  RequiredOutput extends PossibleParameterKeys<OutputParameters>,
+> {
+  private routes: Array<
+    [BlockActionConstraint, BlockActionHandler<typeof this.func.definition>]
+  >;
+
+  constructor(
+    private func: SlackFunction<
+      InputParameters,
+      OutputParameters,
+      RequiredInput,
+      RequiredOutput
+    >,
+  ) {
+    this.func = func;
+    this.routes = [];
+  }
+
+  /**
+   * Add an action handler for something that can match an action event.
+   * @param {BlockActionConstraint} actionConstraint A BlockActionConstraintField(i.e. a string, array of strings or regular expression) or more complex BlockActionConstraintObject to match incoming block action events. A BlockActionConstraintField parameter are matched with a block action event's `action_id` property. A BlockActionConstraintObject parameter allows to match with other block action event properties like `block_id` as well as `action_id`. If multiple properties are specified using BlockActionConstraintObject, then the event must match ALL provided BlockActionConstraintObject properties.
+   * @returns {ActionsRouter}
+   */
+  addHandler(
+    actionConstraint: BlockActionConstraint,
+    handler: BlockActionHandler<typeof this.func.definition>,
+  ): ActionsRouter<
+    InputParameters,
+    OutputParameters,
+    RequiredInput,
+    RequiredOutput
+  > {
+    this.routes.push([actionConstraint, handler]);
+    return this;
+  }
+
+  /**
+   * Returns a method handling routing of action payloads to the appropriate action handler.
+   * The output of export() should be attached to the `blockActions` export of your function.
+   */
+  export(): BlockActionHandler<typeof this.func.definition> {
+    return async (context) => {
+      const action: BlockAction = context.action;
+      const handler = this.matchHandler(action);
+      if (handler === null) {
+        // TODO: what do in this case?
+        // perhaps the user typo'ed the action id when registering their handler or defining their block.
+        // In the local-run case, this warning should be apparent to the user, but in the deployed context, this might be trickier to isolate
+        console.warn(
+          `Received block action payload with action=${
+            JSON.stringify(action)
+          } but this app has no action handler defined to handle it!`,
+        );
+        return;
+      }
+      return await handler(context);
+    };
+  }
+
+  /**
+   * Return the first registered ActionHandler that matches the action ID string provided.
+   */
+  matchHandler(
+    action: BlockAction,
+  ): BlockActionHandler<typeof this.func.definition> | null {
+    for (let i = 0; i < this.routes.length; i++) {
+      const route = this.routes[i];
+      let [constraint, handler] = route;
+      // Handle different constraint types below
+      if (
+        constraint instanceof RegExp || constraint instanceof Array ||
+        typeof constraint === "string"
+      ) {
+        // Normalize simple string constraints to be an array of strings for consistency in handling inside this method.
+        constraint = normalizeConstraintToArray(constraint);
+        // Handle the case where the constraint is either a regex or an array of strings to match against action_id
+        if (matchBlockActionConstraintField(constraint, "action_id", action)) {
+          return handler;
+        }
+      } else {
+        // Assumes an object as a constraint (type BlockActionConstraintObject)
+        // Return first match *within* any of the defined fields on the constaint object, but ensure there is a match on *all* defined fields
+        // Effectively a logical AND across the action_id and block_id field(s)
+        let actionIDMatched = false;
+        let blockIDMatched = false;
+        if (constraint.action_id) {
+          actionIDMatched = matchBlockActionConstraintField(
+            normalizeConstraintToArray(constraint.action_id),
+            "action_id",
+            action,
+          );
+        }
+        if (constraint.block_id) {
+          blockIDMatched = matchBlockActionConstraintField(
+            normalizeConstraintToArray(constraint.block_id),
+            "block_id",
+            action,
+          );
+        }
+        if (
+          (constraint.block_id && constraint.action_id && blockIDMatched &&
+            actionIDMatched) ||
+          (constraint.block_id && blockIDMatched) ||
+          (constraint.action_id && actionIDMatched)
+        ) {
+          return handler;
+        }
+      }
+    }
+    return null;
+  }
+}
+
+function normalizeConstraintToArray(constraint: BlockActionConstraintField) {
+  if (typeof constraint === "string") {
+    constraint = [constraint];
+  }
+  return constraint;
+}
+
+function matchBlockActionConstraintField(
+  constraint: BlockActionConstraintField,
+  field: keyof BlockActionConstraintObject,
+  action: BlockAction,
+) {
+  if (constraint instanceof RegExp) {
+    if (action[field].match(constraint)) {
+      return true;
+    }
+  } else if (constraint instanceof Array) {
+    for (let j = 0; j < constraint.length; j++) {
+      const c = constraint[j];
+      if (action[field] === c) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
--- a/src/functions/routers/action_router_test.ts
+++ b/src/functions/routers/action_router_test.ts
@@ -1,9 +1,119 @@
 import { assertEquals, assertExists, assertMatch } from "../../dev_deps.ts";
 import { BlockActionsRouter } from "./action_router.ts";
+import type { ActionContext, BlockAction } from "./types.ts";
+import type { FunctionRuntimeParameters } from "../types.ts";
+import type {
+  ParameterSetDefinition,
+  PossibleParameterKeys,
+} from "../../parameters/mod.ts";
+import type { SlackFunction } from "../mod.ts";
 import { DefineFunction, Schema } from "../../mod.ts";
-import { DEFAULT_ACTION, SlackActionHandlerTester } from "../tester/mod.ts";
 
-// Test fixtures: a basic function definition and associated block action router to test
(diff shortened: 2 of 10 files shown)


src/functions/routers/mod.ts
```diff
@@ -0,0 +1,69 @@
...
+    func: SlackFunction<InputParameters, OutputParameters, RequiredInput, RequiredOutput>,
+  ) {
+    this.func = func;
+    this.routes = {};
+  }
+
+  /**
+   * Add an action handler for a specific action_id.
+   */
+  add(action_id: string, handler: ActionHandler<typeof this.func.definition.input_parameters>): ActionsRouter<InputParameters, OutputParameters, RequiredInput, RequiredOutput> {
```
filmaj, line 44: TODO: add more signatures to this method. See bolt-js

src/functions/routers/mod.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * Returns a method handling routing of action payloads to the appropriate action handler.
+   * The output of export() should be attached to the `blockActions` export of your function.
+   */
+  export(): ActionHandler<typeof this.func.definition> {
+    return async (context) => {
+      const action = context.action.action_id as string;
+      const handler = this.matchHandler(action);
```
selfcontained, line 83: I think we should expand what our criteria matching handles to both `action_id` and `block_id`. This is the route bolt ended up going, and it ended up satisfying a lot of use-cases. For example, being able to do the following:


### 2 basic ways to call

First way: this assumes arg is an action_id
```ts
router.add(action_id: string, ...)
```

Second Way: This allows for more complex matching against block_ids and action_ids
```ts
router.add(criteria: { block_id?: string, action_id?: string }, ...)
```

Then we can expand where we accept a `string` to accept `regex`, or an `array` of string/regex

### Examples:

When arg is just a string, just matches on action_id (the most common scenario)
```ts
router.add('my_action_id'), ...);
```

Matches on just block_id 
```ts
router.add({ block_id: 'my_block_id'}, ...)
```

Matches on both block_id and action_id
```ts
router.add({ block_id: 'my_block_id', action_id: 'my_action_id'}, ...)
```

Matches on just action_id - mostly for consistency w/ the interface, but you'd probably use the plain string arg for just action_id
```ts
router.add({ action_id: 'my_block_id'}, ...)
```

#### Arrays
This feels like we could defer for v1 and add as a followup unless it's easier to just do now

Match on multiple action ids
```ts
router.add(['action1', 'action2'), ...);
```

Also support arrays in all the other scenarios
```ts
router.add({ block_id: ['block1', 'block2']}, ...)
```
This one feels more complicated, and perhaps less useful but for consistency in the interface would be nice to support, would just match if any block_ids match, and any action_ids match
```ts
router.add({ block_id: ['block1', 'block2'], action_id: ['action1', 'action2']}, ...)
```

###Regex
We could also defer regex and start w/ just strings, but eventually use a regex anywhere a string can be used too
```ts
router.add(/^my_prefix/, ...)
router.add({ block_id: /^my_blocks/, action_id: /_id$/}, ...)
// also support arrays w/ regexes
```

src/functions/routers/mod.ts
```diff
@@ -0,0 +1,125 @@
...
+        console.warn(
+          `Received block action payload with action_id=${action} but this app has no action handler defined to handle it!`,
+        );
+        return;
+      }
+      if (handler.constructor.name === "AsyncFunction") {
+        return await handler(context);
+      } else {
+        return handler(context);
+      }
```
shapirone, line 96: For my own education — is it a problem to await a sync function here? I understand it'll treat the sync function as a promise and push the sync fn execution to a microtask queue, but will that have real impact? 🤔

src/functions/routers/mod.ts
```diff
@@ -0,0 +1,125 @@
...
+> {
+  private func: SlackFunction<
+    InputParameters,
+    OutputParameters,
+    RequiredInput,
+    RequiredOutput
+  >;
+  private routes: Array<
+    Array<string[] | RegExp | ActionHandler<typeof this.func.definition>>
+  >;
```
shapirone, line 43: How this is currently written will allow an array of any amount with no specific ordering. If I'm following correctly [this is intended to be a tuple](https://github.com/slackapi/deno-slack-sdk/pull/57/files#diff-72ea06050e8c947dc58ca4d49cd8fff07eeace6a35b0dd16b24ec1e5ab3b3828R72) where the first item can be an array of strings or a single RegExp.

If that's the case it seems like we could do
```ts
routes: [string[] | RegExp, ActionHandler<typeof this.func.definition>]
```

which will keep us to something like
```ts
this.routes = [['hi', 'fil' , 'how', 'are', 'you'], () => {}];
this.routes = [regExpValue, () => {}];
```

src/functions/routers/mod.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * Return the first registered ActionHandler that matches the action ID string provided.
+   */
+  matchHandler(
+    action: string,
+  ): ActionHandler<typeof this.func.definition> | null {
+    for (let i = 0; i < this.routes.length; i++) {
+      const route = this.routes[i];
```
shapirone, line 107: ultra nit: Assuming each item in `this.routes` is indeed a tuple, we could destructure to give them meaningful names.

```ts
const [constraint, handler] = route;
```

src/functions/routers/mod.ts
```diff
@@ -0,0 +1,125 @@
...
+      if (constraint instanceof RegExp) {
+        if (action.match(constraint)) {
+          return route[1] as ActionHandler<typeof this.func.definition>;
+        }
+      } else {
+        const stringConstraints = constraint as string[];
+        for (let j = 0; j < stringConstraints.length; j++) {
+          const c = stringConstraints[j];
+          if (action === c) {
+            return route[1] as ActionHandler<typeof this.func.definition>;
```
shapirone, line 118: If we can properly type `route[1]` as an `ActionHandler` above when we type `this.routes`, we won't need any of the casting in these lines!

shapirone (changes requested):

src/types.ts
```diff
@@ -11,6 +11,8 @@ import type { ICustomType } from "./types/types.ts";
 // An app manifest is generated based on what this has defined in it
 
 export type {
+  ActionContext,
```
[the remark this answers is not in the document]
filmaj: Ya good point, I followed the lead of the `SlackFunctionHandler` type, which would direct users down the path to typing the entire function handler and thus not needing to type the parameters to that function. Can remove this - will do.

src/types.ts
```diff
@@ -11,6 +11,8 @@ import type { ICustomType } from "./types/types.ts";
...
+  ActionHandler,
```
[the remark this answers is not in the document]
filmaj: Yeah that's a good point. More of a leftover from initial testing; since the initial action registering `add` method didn't support complex matching to action IDs, I had to register it multiple times - thus why I assigned it to a variable. I imagine once I implement the more complex matching that Brad suggests, in practice this wouldn't be needed.

I will tackle the more complex action matching that Brad suggests first, and then based on that, remove exporting any of the underlying types if it's not needed.

src/functions/routers/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  FunctionRuntimeParameters,
+} from "../types.ts";
+
+export type ActionHandler<Definition> = Definition extends
+  FunctionDefinitionArgs<infer I, infer O, infer RI, infer RO> ? {
+    (
+      context:
+        & Omit<FunctionContext<FunctionRuntimeParameters<I, RI>>, "event">
+        & ActionContext<FunctionRuntimeParameters<I, RI>>,
+    ): Promise<void> | void;
```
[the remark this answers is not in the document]
filmaj: Yeah I brought this up with @selfcontained earlier in the week. IIRC we weren't sure what users would do in these action handlers... I think the typical case would be to call one of the function `complete*` callbacks, but that's not necessarily the case (more complex functions might not be completing the function lifecycle in action handlers is how I understand it). I believe @selfcontained was OK with putting the responsibility of calling these complete callbacks themselves - at this point, anyways.

Ultimately, the calling of the function callbacks is done in the deno-slack-runtime library: compare the end of [the `run-block-action` function in there](https://github.com/slackapi/deno-slack-runtime/blob/main/src/run-block-actions.ts) to the [`run-function` function](https://github.com/slackapi/deno-slack-runtime/blob/main/src/run-function.ts) - the latter calls the `function.complete*` APIs, the former does not.

src/functions/routers/mod.ts
filmaj: Probably not, other than linter warnings 🤷 will make this change.

src/functions/routers/mod.ts
filmaj: Haha yeah good call, I struggled with how to type nested arrays with specific types at specific positions of the array 🤪 didn't know you could do what you suggest! Will give it a go.

src/functions/routers/mod.ts
selfcontained: My suggestion would be we follow the path paved from bolt js and make it an AND. From an example on the bolt js docs:

```js
// Your listener function will only be called when the action_id matches 'select_user' AND the block_id matches 'assign_ticket'
app.action({ action_id: 'select_user', block_id: 'assign_ticket' },

```

We're adding support for arrays here, which bolt doesn't have, but I think the AND between block_id and action_id can behave the same way as bolt. What do you think?

filmaj: Ya that sounds good, just will be tricky to communicate that but that can be shelved as a documentation problem 😄 
The way I'm understand it is: logical OR for matching to a specific, single property, but logical AND _across_  all properties of complex constraint objects.

src/functions/routers/types.ts
```diff
@@ -0,0 +1,58 @@
+import {
+  FunctionContext,
+  FunctionDefinitionArgs,
+  FunctionParameters,
+  FunctionRuntimeParameters,
+} from "../types.ts";
+
+export type ActionHandler<Definition> = Definition extends
```
selfcontained, line 8: nit - could we consider naming this `BlockActionHandler` (and a few other Action* types) for consistency?

src/functions/routers/mod.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (constraint.block_id) {
+          blockIDMatched = matchBlockActionConstraintField(
+            normalizeConstraintToArray(constraint.block_id),
+            "block_id",
+            action,
+          );
+        }
+        if (
+          (constraint.block_id && constraint.action_id && blockIDMatched &&
+            actionIDMatched) ||
```
shapirone, line 138: As discussed, this check doesn't ensure that both `block_id` and `action_id` match since the below checks will pass if either matches.

src/functions/routers/types.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+// TODO: all below stolen from deno-slack-runtime, need to centralize these somewhere
+export type BlockActionInvocationBody<
+  InputParameters extends FunctionParameters,
+> = {
+  type: "block_actions";
+  actions: BlockAction[];
+  bot_access_token?: string;
+  function_data?: FunctionData<InputParameters>;
```
selfcontained, line 30: In the runtime I left this as optional for some backwards compatibility - as it's not always present until a full rollout happens. I believe we can treat it always present here, as the handler would never be invoked if it wasn't there. It would simplify how you pull out properties from here in your handler code a little, like getting the execution_id.

src/functions/routers/action_router_test.ts
```diff
@@ -0,0 +1,355 @@
...
+  source_file: "whatever",
+  input_parameters: {
+    properties: {
+      garbage: { type: Schema.types.string },
+    },
+    required: [],
+  },
+});
+const { createContext } = SlackActionHandlerTester(func);
+const inputs = { garbage: "in, garbage out" };
```
shapirone, line 19: I think we should test for the case of accessing inputs from within a handler. In this case, we should have `ctx.inputs.garbage` available to us. This totally works, but figure it's good to add a test and make sure it continues to work 😅

src/functions/routers/action_router.ts
```diff
@@ -0,0 +1,191 @@
...
+    this.func = func;
+    this.routes = [];
+  }
+
+  /**
+   * Add an action handler for something that can match an action event.
+   * @param {BlockActionConstraint} actionConstraint A BlockActionConstraintField(i.e. a string, array of strings or regular expression) or more complex BlockActionConstraintObject to match incoming block action events. A BlockActionConstraintField parameter are matched with a block action event's `action_id` property. A BlockActionConstraintObject parameter allows to match with other block action event properties like `block_id` as well as `action_id`. If multiple properties are specified using BlockActionConstraintObject, then the event must match ALL provided BlockActionConstraintObject properties.
+   * @returns {ActionsRouter}
+   */
+  add(
```
shapirone, line 76: nit: I'm wondering if we should give some additional thought on how this method is named. Calling `BlockActionsRouter(...).add(...)` isn't too clear to me on what I'm adding. `BlockActionsRouter(...).addHandler(...)` is more clear, or alternatively maybe `BlockActionHandlerRouter(...).add(...)`? Just thinking out loud here

src/functions/routers/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  body: BlockActionInvocationBody<InputParameters>;
+  action: BlockAction;
+};
+
+// TODO: all below stolen from deno-slack-runtime, need to centralize these somewhere
+export type BlockActionInvocationBody<
+  InputParameters extends FunctionParameters,
+> = {
+  type: string;
+  actions: BlockAction[];
```
shapirone, line 30: ❓ This is an array of Block Actions being passed to an individual action handler. Do you know what this will actually look like? Is it just all possible Block Actions for the fn?

src/functions/routers/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  bot_access_token?: string;
```
shapirone, line 31: We're exposing this as `token` at the top level of the handler context. In the default function definition we don't type the `bot_access_token` even though it comes in through `body.event`. Is there a reason we want to here?

src/functions/tester/types.ts
```diff
@@ -62,3 +66,52 @@ export type SlackFunctionTesterFn = {
...
+};
+
+export type SlackActionHandlerTesterResponse<
+  InputParameters extends ParameterSetDefinition,
+  RequiredInput extends PossibleParameterKeys<InputParameters>,
+> = {
+  createContext: CreateActionHandlerContext<InputParameters, RequiredInput>;
+};
+
+// Slack Action Handler Tester sux
```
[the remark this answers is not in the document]
filmaj: Oh my god I think I left this comment in there at the height of my typescript troubles / frustrations 🤦

src/functions/tester/mod.ts
```diff
@@ -4,9 +4,22 @@ import {
...
-import { CreateContext, SlackFunctionTesterFn } from "./types.ts";
+import {
+  CreateActionHandlerContext,
+  CreateFunctionContext,
+  SlackActionHandlerTesterFn,
+  SlackFunctionTesterFn,
+} from "./types.ts";
 
 export const DEFAULT_FUNCTION_TESTER_TITLE = "Function Test Title";
+export const DEFAULT_ACTION = {
```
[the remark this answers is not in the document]
filmaj: They're not used but they certainly are part of the payloads received... should we therefore type them?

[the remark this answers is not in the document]
shapirone: Maybe they are part of the payload depending upon the `type`? I.e. `text` and `style` are valid for `type: button`. Maybe just adding a `[key: string]: any` to the `BlockAction` type will cover it?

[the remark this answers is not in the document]
filmaj: Yeah I think I will do that for now... I was looking at webapp lambda dispatch code to see how payloads are build up in webapp before dispatching out to lambda for execution.. it's... a lot.

src/functions/routers/types.ts
```diff
@@ -0,0 +1,60 @@
...
+type ActionSpecificContext<InputParameters extends FunctionParameters> = {
+  body: BlockActionInvocationBody<InputParameters>;
+  action: BlockAction;
+};
+
+// TODO: all below stolen from deno-slack-runtime, need to centralize these somewhere
+export type BlockActionInvocationBody<
+  InputParameters extends FunctionParameters,
+> = {
+  type: string;
```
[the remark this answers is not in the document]
filmaj: Oh man I'm 10 levels deep trying to figure out in webapp where this property gets set 🤣 😢 been trying to dig that up for an hour now. Arghhhhh

src/functions/routers/action_router.ts
filmaj: I like `addHandler`. I will rename it.

src/functions/routers/types.ts
shapirone: Personally, I don't think we should surface this from the SDK. This follows the pattern we have for the primary function, and it might be confusing for devs to see two different tokens

src/functions/routers/types.ts
```diff
@@ -0,0 +1,61 @@
...
+  FunctionDefinitionArgs,
+  FunctionParameters,
+  FunctionRuntimeParameters,
+} from "../types.ts";
+
+export type BlockActionHandler<Definition> = Definition extends
+  FunctionDefinitionArgs<infer I, infer O, infer RI, infer RO> ? {
+    (
+      context: ActionContext<FunctionRuntimeParameters<I, RI>>,
+    ): Promise<void> | void;
```
shapirone, line 12: In the runtime we have this return `any`, should we do that here as well?

src/functions/routers/types.ts
shapirone: I think I accidentally marked this comment resolved, but I agree with @selfcontained that it should be treated as always present here. We need it in order to complete the function via 
```ts
client.functions.completeSuccess({
    outputs: {},
    function_execution_id: body.function_data?.execution_id,
  });
```

src/functions/tester/mod.ts
```diff
@@ -64,3 +78,42 @@ export const SlackFunctionTester: SlackFunctionTesterFn = <
 
   return { createContext };
 };
+
+export const SlackActionHandlerTester: SlackActionHandlerTesterFn = <
```
shapirone, line 82: I find it a bit odd that this internal only test helper lives here in `functions/tester` which is intended for devs to test their functions. I'd like to either move this somewhere else specifically for our tests, or add a comment calling out this is internal only and therefore dissimilar to `SlackFunctionTester` above

shapirone (changes requested):

src/functions/tester/mod.ts
filmaj: For now, I'll move it into the action router tests, since that's the only place it is used, and we can revisit if/how we want to expose it to devs down the road.

shapirone (approved):
letsdothis

PR #58, opened 2022-07-12
Enable codecov coverage reporting.

filmaj:
Ignore functions.json schema from linting or checking.

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -1,12 +1,14 @@
 {
   "fmt": {
     "files": {
-      "include": ["src"]
+      "include": ["src"],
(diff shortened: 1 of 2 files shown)


deno.jsonc
```diff
@@ -1,12 +1,14 @@
 {
   "fmt": {
     "files": {
-      "include": ["src"]
+      "include": ["src"],
+      "exclude": ["src/schema/slack/functions/_scripts/functions.json"]
```
shapirone, line 5: Why do we want to exclude this? The file isn't checked into git, so this would just be running `fmt` locally right?

shapirone (approved):

PR #62, opened 2022-07-20
Initial implementation of the slack logger

shorbu:
###  Summary
Currently stdout/stderr of AWS lambdas is being parsed by webapp at the end of execution for specific strings in order to generate structured/filterable/searchable logs (`apps.activities.list`). 

This helper is being provided to avoid having Slack app devs bother with the intricate format expectations of their logging.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -8,37 +8,37 @@ enum LogLevel {
 }
 
 export function debug<T>(
-  msg: (T extends (...args: any[]) => any ? never : T) | (() => T),
+  msg: (T extends (...args: unknown[]) => unknown ? never : T) | (() => T),
 ): string {
   return write(msg, LogLevel.DEBUG);
 }
 
 export function info<T>(
-  msg: (T extends (...args: any[]) => any ? never : T) | (() => T),
+  msg: (T extends (...args: unknown[]) => unknown ? never : T) | (() => T),
 ): string {
   return write(msg, LogLevel.INFO);
 }
 
 export function warn<T>(
-  msg: (T extends (...args: any[]) => any ? never : T) | (() => T),
+  msg: (T extends (...args: unknown[]) => unknown ? never : T) | (() => T),
 ): string {
   return write(msg, LogLevel.WARN);
 }
 
 export function error<T>(
-  msg: (T extends (...args: any[]) => any ? never : T) | (() => T),
+  msg: (T extends (...args: unknown[]) => unknown ? never : T) | (() => T),
 ): string {
   return write(msg, LogLevel.ERROR);
 }
 
 export function fatal<T>(
-  msg: (T extends (...args: any[]) => any ? never : T) | (() => T),
+  msg: (T extends (...args: unknown[]) => unknown ? never : T) | (() => T),
 ): string {
   return write(msg, LogLevel.FATAL);
 }
 
 function write<T>(
-  msg: (T extends (...args: any[]) => any ? never : T) | (() => T),
+  msg: (T extends (...args: unknown[]) => unknown ? never : T) | (() => T),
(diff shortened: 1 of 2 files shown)


src/logger.ts
```diff
@@ -0,0 +1,80 @@
...
+  } else {
+    logMessage = toString(msg);
+  }
+
+  const formattedLog =
+    `{"loggerName":"slack","msg":"${logMessage}","level":${level},"datetime":"${
+      new Date().toISOString()
+    }"}`;
+
+  console.log(formattedLog);
```
filmaj, line 59: Could we, based on the `LogLevel` passed in, direct to the appropriate underlying [Deno `console` method](https://doc.deno.land/deno/stable/~/Console)? Something like, `DEBUG` => `console.debug`, `INFO` => `console.info`, `WARN` => `console.warn`, `ERROR` => `console.error`. Not sure about `FATAL` or `NONE` - no relevant level there that's obvious to me.

src/logger.ts
```diff
@@ -0,0 +1,80 @@
...
+  let logMessage: string;
+
+  if (msg instanceof Function) {
+    res = msg();
+    logMessage = toString(res);
+  } else {
+    logMessage = toString(msg);
+  }
+
+  const formattedLog =
```
filmaj, line 54: I wonder if we can conditionally apply this formatting only if we are running in a hosted Slack Cloud context.. Maybe detecting for a [reserved AWS Lambda environment variable](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime) via [`Deno.env`](https://doc.deno.land/deno/stable/~/Deno.env)? Or maybe the deno library running on Lambda that does the dispatching injects an environment variable to make that explicit.

The reason I ask is because when I used this logger and ran locally, the output was quite hard to read.

filmaj (changes requested):
Overall looks fine to me.

My main concern with the implementation as-is is that the logs look a lot different when running locally (via `slack run`) vs. when running remotely and logs being pulled down (via `slack deploy` and `slack activity`).

Logs as displayed in user terminal via `slack activity`:

```
2022-07-22 14:18:44 [info] [Fn03RCHQHK5E] (Trace=Tr03QRC4HA4C) [2022-07-22T18:18:44.449Z] reversing yo  yo yo.
2022-07-22 14:18:44 [info] [Fn03RCHQHK5E] (Trace=Tr03QRC4HA4C) [2022-07-22T18:18:44.449Z] SLACK_API_URL=undefined
```

Logs as displayed when running locally via `slack run`:

```
{"loggerName":"slack","msg":"reversing the wind below.","level":20,"datetime":"2022-07-22T18:19:24.432Z"}
{"loggerName":"slack","msg":"SLACK_API_URL=https://slack.com/api/","level":20,"datetime":"2022-07-22T18:19:24.433Z"}
```

The locally-running logs are hard to read for users as its (intentionally) meant for machine consumption, not user consumption. I wonder if we can detect whether we are in a locally running context, and simply forward the arguments of the methods to the appropriate `console` methods?

src/logger.ts
shorbu: 🤔 This is interesting, while I agree that it makes the stdout from `slack run` much more complex to read. 
I believe that since the `slack run` still executes the function in AWS, we wouldn't be able to generate activities as expected and someone calling `apps.activities.list` for the "(dev)" app wouldn't have the same result as the `slack deploy` version of the app.

I'm not sure if this matters, or if we're comfortable with that...

shorbu:
After discussing with @curtisallen, @monica and @mcodik  we decided to post-pone this in favour of a future cleaner solution (i.e. overriding console.log outside of the user-code layer).

PR #67, opened 2022-07-25
Resolves type errors

srajiang:
###  Summary

These type errors show up when trying to build the SDK. Errors are related to referencing private `this` in the return type of a public method in an exported class.

image

### DIFF
--- a/src/functions/routers/action_router.ts
+++ b/src/functions/routers/action_router.ts
@@ -3,6 +3,7 @@ import {
   PossibleParameterKeys,
 } from "../../parameters/mod.ts";
 import { SlackFunction } from "../mod.ts";
+import { FunctionDefinitionArgs } from "../types.ts";
 import type {
   BlockAction,
   BlockActionConstraint,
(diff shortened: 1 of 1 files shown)


src/functions/routers/action_router.ts
```diff
@@ -75,7 +76,14 @@ export class ActionsRouter<
    */
   addHandler(
```
shapirone, line 77: Is there a reason we're leaving this in when we type `routes` on line 56? Still looks like

```ts
private routes: Array<
    [BlockActionConstraint, BlockActionHandler<typeof this.func.definition>]
  >;
```

shapirone (commented):
Excellent! I tested this and it looks like our typeahead is still working within the `addHandler` method. I think we missed one case, but maybe that was intentional?

src/functions/routers/action_router.ts
srajiang: @shapirone - Yes, TS seems to have no complains with using `this`, I guess because in the other cases we were referencing `this` in the return types of public methods. I think we can leave this alone.

shapirone (approved):

neptunel (approved):
Looking great, thanks @srajiang

PR #68, opened 2022-07-27
Adding some typing for block_actions body

selfcontained:
###  Summary

This adds a little bit more typing for the `body` of the BlockActionsRouter handlers. I've added the new `interactivity` portion of the payload, and then flushed out some of the other more common and simple properties for now, such as `user` and `team`, as well as other portions of the payload. Ideally we can come up with a strategy to centralize these payload types to share across our libraries.

### DIFF
--- a/src/functions/routers/types.ts
+++ b/src/functions/routers/types.ts
@@ -42,11 +42,13 @@ type FunctionData<InputParameters extends FunctionParameters> = {
 };
 
 type Interactivity = {
-  interactor: {
-    secret: string;
-    id: string;
-  };
+  interactor: UserContext;
   interactivity_pointer: string;
+};
+
+type UserContext = {
+  secret: string;
+  id: string;
 };
 
 export type BlockActionConstraint =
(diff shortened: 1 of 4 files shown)


src/functions/routers/types.ts
```diff
@@ -42,12 +41,12 @@ type FunctionData<InputParameters extends FunctionParameters> = {
...
-  type: string;
-  block_id: string;
-  action_id: string;
-  // deno-lint-ignore no-explicit-any
-  [key: string]: any;
+type Interactivity = {
+  interactor: {
+    secret: string;
+    id: string;
+  };
```
shapirone, line 48: nit: Should we have a user context type that we set interactor to, or do you think that's not valuable since we're (currently) not surfacing user context separately?

I'm currently thinking that we'll probably want a type for user context at some point and think that this should be tied to it whenever we do.

shapirone (approved):
Looks great, just a nit about creating a separate UserContext type and using it in Interactivity

src/functions/routers/types.ts
selfcontained: Possibly, I'll go ahead and break it out for now, even for clarity it might be nice. I'm not sure I see an immediate payload that would include just that piece, but still a good idea.

PR #69, opened 2022-07-27
Use Custom Types for `user_context` and `interactivity` types

shapirone:
###  Summary

This uses Custom Slack Types instead of the hardcoded checks!

#### Testing
Using `user_context` and `interactivity` as function input/output parameters within a workflow should continue to work and not change

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -13,10 +13,6 @@ import {
 import type SchemaTypes from "../schema/schema_types.ts";
 import type SlackSchemaTypes from "../schema/slack/schema_types.ts";
 import { SlackManifest } from "../manifest/mod.ts";
-import {
-  InteractivityParameterDefinition,
-  UserContextParameterDefinition,
-} from "../schema/slack/schema_types.ts";
 
 export type { BlockActionHandler } from "./routers/types.ts";
 
@@ -56,14 +52,6 @@ type FunctionInputRuntimeType<Param extends ParameterDefinition> =
       ? Param extends TypedObjectParameterDefinition
         ? TypedObjectFunctionInputRuntimeType<Param>
       : UnknownRuntimeType
-    : Param["type"] extends typeof SlackSchemaTypes.interactivity
-      ? TypedObjectFunctionInputRuntimeType<
-        typeof InteractivityParameterDefinition
-      >
-    : Param["type"] extends typeof SlackSchemaTypes.user_context
-      ? TypedObjectFunctionInputRuntimeType<
-        typeof UserContextParameterDefinition
-      >
     : Param["type"] extends
       | typeof SlackSchemaTypes.user_id
       | typeof SlackSchemaTypes.usergroup_id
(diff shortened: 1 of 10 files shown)


shapirone:
@selfcontained the second commit in this PR introduces some optional structure changes that might set us up for the future. Let me know if you'd prefer the simpler approach for now without that additional structure.

src/schema/slack/types/mod.ts
```diff
@@ -0,0 +1,10 @@
+const SlackTypes = {
```
shapirone, line 1: It might be a little confusing that this is called `SlackTypes` here, when throughout our code base we import `SlackTypes` from `src/schema/slack/schema_types.ts` which will be the combination of these slack primitives and the slack custom types. The only case where we import this is in `src/parameters.ts/types.ts`

src/test_utils.ts
```diff
@@ -0,0 +1,12 @@
+import { assertEquals } from "./dev_deps.ts";
+
+// deno-lint-ignore ban-types
+type IsAny<T> = unknown extends T ? T extends {} ? T : never : never;
+type NotAny<T> = T extends IsAny<T> ? never : T;
+
+/** @description Prevents a value of type `any` from being passed into `assertEquals` */
+export const assertEqualsTypedValues = <T>(
```
shapirone, line 8: Open to a better name here

src/test_utils.ts
```diff
@@ -0,0 +1,12 @@
+import { assertEquals } from "./dev_deps.ts";
+
+// deno-lint-ignore ban-types
+type IsAny<T> = unknown extends T ? T extends {} ? T : never : never;
```
shapirone, line 4: We could likely leverage this type again for another test util (`assertEqualsAnyTape`) to verify that we're getting `any` when we actually expect it to meet that `UnknownRuntimeType`.

src/functions/types.ts
```diff
@@ -39,11 +36,36 @@ export type FunctionInvocationBody = {
...
  * @description Maps a ParameterDefinition into a runtime type, i.e. "string" === string.
  */
-type FunctionInputRuntimeType<Param extends ParameterDefinition> =
-  Param["type"] extends typeof SchemaTypes.string ? string
+type FunctionInputRuntimeType<
+  Param extends ParameterDefinition,
+  CurrentDepth extends RecursionDepthLevel = 0,
+> =
+  // Recurse through Custom Types, stop when we hit our max depth
+  CurrentDepth extends MaxRecursionDepth ? UnknownRuntimeType
```
shapirone, line 62: Without introducing this recursion depth check, we hit the `Type instantiation is excessively deep and possibly infinite` type error in `SlackFunctionTester`

selfcontained: This is some black magic voodooshit

src/test_utils.ts
selfcontained: it's perfect

selfcontained (approved):
Great job finding a way to make custom types just "fit in". shipit

PR #70, opened 2022-07-29
Default tokenManagementEnabled to false for runOnSlack apps

srajiang:
###  Summary

`Manifest` now outputs `token_management_enabled: false` by default for all run on slack apps (aka where function_runtime is set as `slack`.

#### Developer Experience
* `tokenManagementEnabled` property does not appear in typeahead. In future, all run on slacks apps created should default to false in the backend so I figured it'd be better to keep this property out of sight. 

https://user-images.githubusercontent.com/55667998/181822386-6986a48f-37a7-4584-ab59-c5ac1ab487df.mov

* Manifest outputs the proper property

https://user-images.githubusercontent.com/55667998/181822461-28baa01e-2292-4441-8157-ea2991b92444.mov

* This does not impact remote / Bolt apps. tokenManagementFalse remains an [optional property f](https://github.com/slackapi/deno-slack-sdk/blob/5c61e062a0da51a30ab6ba0ce0378a31f93a8b0d/src/manifest/types.ts#L63)or these apps. 

### DIFF
--- a/src/manifest/mod.ts
+++ b/src/manifest/mod.ts
@@ -231,6 +231,10 @@ export class SlackManifest {
   private assignRunOnSlackManifestProperties(manifest: ManifestSchema) {
     const def = this.definition as ISlackManifestRunOnSlack;
 
+    // Run on Slack Apps do not manage access tokens
+    // This is set by default as false
+    manifest.oauth_config.token_management_enabled = false;
+
     // External Auth providers
     if (def.externalAuthProviders) {
       manifest.external_auth_providers = def.externalAuthProviders?.reduce(
--- a/src/manifest/manifest_test.ts
+++ b/src/manifest/manifest_test.ts
@@ -214,6 +214,38 @@ Deno.test("Manifest() properly converts callback_id to proper key", () => {
   assertEquals(manifest.types, { "Using Callback": { type: "boolean" } });
 });
 
+Deno.test("Manifest() always sets token_management_enabled to false for runOnSlack: true apps", () => {
+  // When runOnSlack is explicitly specified as true, token_management_enabled must be set to false
+  const definition: SlackManifestType = {
+    runOnSlack: true,
+    name: "",
+    description: "",
+    backgroundColor: "#FFF",
+    longDescription: "",
+    displayName: "",
+    icon: "",
+    botScopes: [],
+  };
+  const manifest = Manifest(definition);
+  assertEquals(manifest.oauth_config.token_management_enabled, false);
+});
+
+Deno.test("Manifest() always sets token_management_enabled to false for function_runtime: slack apps", () => {
+  // SlackManifestType will default to function_runtime == slack when runOnSlack property omitted
+  // AND when no remote-only features are specified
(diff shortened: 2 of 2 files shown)


src/manifest/mod.ts
```diff
@@ -231,6 +231,10 @@ export class SlackManifest {
   private assignRunOnSlackManifestProperties(manifest: ManifestSchema) {
     const def = this.definition as ISlackManifestRunOnSlack;
 
+    // Run on Slack Apps do not manage access tokens
+    // This is set by default as false
+    manifest.oauth_config.token_management_enabled = false;
```
selfcontained, line 236: What do you think about adding a check to see if it isn't explicitly set to `true`, and if not, then set it to false? I don't have a good reason to support the need for someone setting this to `true` for a hosted app, and it shouldn't be, but in the vein of allowing for the developer to be in control at a lower level w/ manifest.json file layering, should we allow it?

selfcontained (approved):
Works as described, thanks for those great videos too! I left one non-blocking comment to consider, but am happy either way.

src/manifest/mod.ts
srajiang: @selfcontained 

I've been reflecting a little in working with these types about the balance between needed and helpful inconsistency between the ManifestSchema and what the developer sees, and making a system that doesn't obfuscate what happens under the hood needlessly giving devs visibility if not control @ a lower level. 

At the moment, devs using typeahead in RunOnSlack apps right now don't even see `tokenManagementEnabled` as an option, largely for the same historical reasons why they don't see manifest.OAuthConfig as an option, and the types actually complain if they try to explicitly set them, so I think we've made it pretty difficult for them to set it to `true`.

The deno sdk seems to have already lifted other oauth configuration specifics like `botScopes` the top level manifest definition since early days. I'm guessing the reason for that is just because at the time things were simpler, it was easier to just put it in top level, token_management_enabled was just a twinkle in Jake's eyes, so it ended up like that. 

It feels like developers who are writing apps shouldn't for the foreseeable future have the option to set this to `true` - otherwise I assume something goes wonky in the backend if they want to still use the hosted solution. Since there's a path to making the option visible to them if it ever does become relevant, by adding the property to the shared types, I think we will be okay continuing without the check!

jpbrabec: > Wrt to your specific question around setting the value to true. It seems to me RoS app devs shouldn't for the foreseeable future have the option to set this to true - otherwise I assume something goes wonky in the backend if they want to still use the hosted solution. I wouldn't want to make the property settable only to lead devs into some kind of failure scenario.

I believe the plan is that it will be possible to have your app use both 1.0 features *and* hosted RoS features, although it's a little awkward for sure. I don't necessarily think it's a blocker for open beta, though. 

> token_management_enabled was just a twinkle in Jake's eyes

I would describe this feature more as "existential dread" than a twinkle 😅

shapirone (approved):
LGTM!

sopokuagyemang (approved):

PR #71, opened 2022-07-29
Adds RoS App Settings defaults: org_deploy_enabled

srajiang:
###  Summary

Adds org_deploy_enabled: true for run on slack apps.

#### Developer Experience
* RoS App devs do not need to set this value, at the moment they also should not be able to set this value either, since there's no option but `true`.
* Remote App devs retain the ability to set this value via manifest.settings.org_deploy_enabled

https://user-images.githubusercontent.com/55667998/181834459-66048168-b96f-468d-a253-2c45f40d2bea.mov

### DIFF
--- a/src/manifest/manifest_test.ts
+++ b/src/manifest/manifest_test.ts
@@ -371,7 +371,7 @@ Deno.test("Manifest() correctly assigns app_directory properties", () => {
   );
 });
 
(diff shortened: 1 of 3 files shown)


selfcontained (approved):
Works as described - thanks for the tests too.

PR #73, opened 2022-08-01
Add support for SendForm and update generation script

shapirone:
###  Summary
The main goal of the PR is to add built-in support for `SendForm`. This PR makes `Schema.slack.functions.SendForm` available and provides the typing for it. Unfortunately, at workflow authoring time we do not provide typing to support object properties, so a developer will need know how to define `fields`.

Since `SendForm` is the first built-in function that uses Custom Types (`interactivity`), this PR also introduces support for Custom Types in our built-in function generation script.

#### SendForm
For a simple test, create a new app from the `workflows` branch of `deno-reverse-string` and update from the `"slack#/functions/send_form"` string to the `Schema.slack.functions.SendForm` reference, then deploy your app and test!

```bash
slack create -t slackapi/deno-reverse-string -b workflows
```

For your own use case, here is the [Tech Spec](https://corp.quip.com/nskyAYDPdVi8/Forms-Built-in-Function-Tech-Spec). **One significant requirement that's easy to miss is that interactivity must either be passed from the trigger, or from the directly preceding step (assuming that step has an interactivity handler).**

#### Built-in Script
Run `./src/schema/slack/functions/_scripts/generate` to see the output files and validate that there are no types being set to `undefined` (`type: undefined`) in the generated files.
_Note that you will see some changes to `SendDM` and `SendMessage`, but those are intentionally excluded from this PR._


src/schema/slack/types/custom/custom_slack_types_test.ts
```diff
@@ -0,0 +1,10 @@
+import { assertEquals } from "../../../../dev_deps.ts";
```
shapirone, line 1: Where to put this test file was a little awkward — do we place it near the Custom Slack Types that worry about this or do we place it near the `DefineType` definition and import the Custom Slack Types? I went with here thinking that if additional tests needed to be added to Custom Slack Types, it'd be easier to find.

src/schema/slack/types/custom/mod.ts
```diff
@@ -1,9 +1,12 @@
...
-const CustomSlackTypes = {
+export const CustomSlackTypes = {
   interactivity: InteractivityType,
   user_context: UserContextType,
 };
 
-export { CustomSlackTypes };
+export const InternalSlackTypes = {
+  form_input_object: FormInput,
+};
```
shapirone, line 12: We want this separate so that devs don't get this as typeahead on `Slack.schema.functions` since it's not to be used as a param in their functions/workflows

src/types/mod.ts
```diff
@@ -30,7 +30,7 @@ export class CustomType<Def extends CustomTypeDefinition>
   }
 
   private generateReferenceString() {
-    return `#/types/${this.id}`;
+    return this.id.startsWith("slack#/") ? this.id : `#/types/${this.id}`;
```
selfcontained, line 33: I think this is fine, but this reminded me that I solved this slightly differently for function reference strings. There I just did a check to see if the id contained a `"#"` - and if it did, assumed it was a full reference and used it, otherwise built the "local" reference string.

```ts
// ensures the function reference is a full path - local functions will only be passing in the function callback id
    this.functionReference = functionReference.includes("#/")
      ? functionReference
      : `${localFnPrefix}${functionReference}`;
```

It may make no difference, but I thought it may be useful if we ever allowed referencing other app's entities at some point, then things would potentially "just work"... or something.

selfcontained (approved):
These changes look great - tested locally as well. One non-blocking comment.

src/types/mod.ts
shapirone: Ah yeah good call! Just going to say this out loud — we may have a problem in the future where we auto-register types from other apps, depending on how we do that. If that will look like `type: A0123/types/example`, no issue. If we're actually importing the `ICustomType` from their app (for typeahead purposes maybe), then we'll need to add logic to handle that. Ex.

```ts
// manifest.ts
import { IncidentType } from "deno-slack-app-repository/incident_app.ts";

DefineFunction({
  ...,
  input_parameters: {
    properties: { incident: { type: IncidentType },
    ...,
  }
});

```

PR #77, opened 2022-08-11
adding slack#/types/date

selfcontained:
###  Summary

Adds `slack#/types/date` to the `Schema`. At runtime it's a string (it's format is `YYYY-MM-DD`).

### Testing
To test this you can use an Open Form step and add a field of `Schema.slack.types.date` - it should render a date-picker like in the screenshot.

image

You can embed that output into a Send a message step's `message` input and you should see the date string you selected, like so:

image

### DIFF
--- a/src/functions/types.ts
+++ b/src/functions/types.ts
@@ -81,7 +81,8 @@ type FunctionInputRuntimeType<
     : Param["type"] extends
       | typeof SlackSchemaTypes.user_id
       | typeof SlackSchemaTypes.usergroup_id
-      | typeof SlackSchemaTypes.channel_id ? string
+      | typeof SlackSchemaTypes.channel_id
+      | typeof SlackSchemaTypes.date ? string
     : Param["type"] extends typeof SlackSchemaTypes.timestamp ? number
     : UnknownRuntimeType;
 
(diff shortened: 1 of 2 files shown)


shapirone:
Do we have an example of what it looks like to set this value, either as a default or when as a parameter passed to a function? Is it just the below?
```ts
MyWorkflow.addStep(FnThatAcceptsDate, {
  date: "2022-08-19"
});
```

selfcontained:
> Do we have an example of what it looks like to set this value, either as a default or when as a parameter passed to a function? Is it just the below?
> 
> ```ts
> MyWorkflow.addStep(FnThatAcceptsDate, {
>   date: "2022-08-19"
> });
> ```
Yup, I was able to default my form's date like that:
```ts
{
  name: "when",
  title: "When",
  type: Schema.slack.types.date,
  default: "2022-08-09",
}
```

shapirone (approved):
Happy path works! Using `date `in `OpenForm` renders the correct field type. Referencing the value properly prints the string.

Found a bug where if we pass a bad value at workflow authoring time to the date input of a function (i.e. `dateField: "im_a_bad_value"`), we don't catch it at manifest validation and the function fails within the workflow engine with no output to the console. We should fix this on the BE, no changes needed here

PR #78, opened 2022-08-15
Exposing team_id to FunctionContext

filmaj:
`team_id` should now be available on all dispatched events to userland functions via runtime.


selfcontained (approved):
Nice!

PR #80, opened 2022-08-17
Add Rich Text to SDK

sbarksdale97:
###  Summary
Adding support for `RichTextValue` within the SDK. 

Previously we had to use `slack#/types/rich_text` to reference the Rich Text type but this changes allows us to use `Schema.slack.types.rich_text`


src/functions/types.ts
```diff
@@ -82,7 +82,8 @@ type FunctionInputRuntimeType<
       | typeof SlackSchemaTypes.user_id
       | typeof SlackSchemaTypes.usergroup_id
       | typeof SlackSchemaTypes.channel_id
-      | typeof SlackSchemaTypes.date ? string
+      | typeof SlackSchemaTypes.date
+      | typeof SlackSchemaTypes.rich_text ? string
```
shapirone, line 86: Could we set `rich_text` to `UnknownRuntimeType` instead of `string`? This will treat it as `any`

shapirone (changes requested):

src/functions/types.ts
selfcontained, line 86: we should probably have this resolve to the `UnknownRuntimeType` (which is an `any`) - since it's a complex object at runtime and not a string. I believe you can just remove it from the checks here and it would fall into that condition below on line 88.

selfcontained (changes requested):
Just one change for the runtime type

src/functions/types.ts
sbarksdale97: Should we keep date as `typeof SlackSchemaTypes.date ? string`? I didn't realize how the ? operator is working here

sbarksdale97: ah are we saying to not include the entire line or to have the line `| typeof SlackSchemaTypes.rich_text ? UnknownRuntimeType`

selfcontained: Good point, it would probably be better to be explicit here, so it would be more like adding a line at the end so we have another condition that matches specifically for rich text. So the block looks something like:

```ts
: Param["type"] extends
  | typeof SlackSchemaTypes.user_id
  | typeof SlackSchemaTypes.usergroup_id
  | typeof SlackSchemaTypes.channel_id
  | typeof SlackSchemaTypes.date ? string
: Param["type"] extends typeof SlackSchemaTypes.timestamp ? number
: Param["type"] extends typeof SlackSchemaTypes.rich_text ? UnknownRuntimeType
: UnknownRuntimeType;
```

selfcontained (approved):

PR #81, opened 2022-08-18
View router discrete methods

filmaj:
# API / Design

Tip of the 🎩  to @selfcontained for suggesting this API.

```typescript
export const { viewSubmission, viewClosed } = ViewsRouter(MyFunctionDefinition)
  .addSubmissionHandler("view_1", async ({ inputs, body /* of type ViewSubmissionInvocationBody */, view, env, token }) => { ... })
  .addClosedHandler("view_1", async ({ inputs, body /* of type ViewClosedInvocationBody */, view, env, token }) => { ... })
```

# Things That Do Not Work

While the deno-slack-sdk compiles, userland code has a problem with the `export const { viewSubmission, viewClosed }` line; it doesn't like the `viewSubmission` and `viewClosed` properties, thinks they don't belong there:

```
error: TS2339 [ERROR]: Property 'viewSubmission' does not exist on type 'ViewRouter<{ requester_id: { type: "slack#/types/user_id"; description: string; }; approval_channel_id: { type: "slack#/types/channel_id"; description: string; }; subject: { type: "string"; description: string; }; details: { ...; }; }, { ...; }, ("requester_id" | ... 2 more ... | "details")[], ("message_ts" | ... 1 ...'.
export const { viewSubmission, viewClosed } = ViewsRouter(ApprovalFunction)
               ~~~~~~~~~~~~~~
    at file:///Users/fmaj/src/interactive-approval/functions/approval/mod.ts:130:16
```

~The above only happens when I attached handlers to the `ViewsRouter` object via the `add*` method(s). If I drop those, then the complaint goes away. This tells me something in the return value from the type-asserted exportedHandler duck-punched `addSubmissionHandler` and `addClosedHandler` magic code in the `ViewsRouter` method is off.~

*☝️This is fixed now - but leaving the explanation for historical purposes. We ended up adjusting how we wrap/export the `viewClosed` and `viewSubmission` handlers so they're directly on the `ViewsRouter` class now.*

# Things I Do Not Like

I feel like I'm duplicating a lot of code for no reason, and that by using generics and TypeScript "properly" I could avoid that. I tried [very](https://github.com/slackapi/deno-slack-sdk/tree/view-router) [much](https://github.com/slackapi/deno-slack-sdk/tree/view-router-base-viewhandler-type) [in](https://github.com/slackapi/deno-slack-sdk/tree/view-router-evt-type-overloaded-as-any) [different](https://github.com/slackapi/deno-slack-sdk/tree/view-router-evt-type-param-overloaded) [ways](https://github.com/slackapi/deno-slack-sdk/tree/view-router-generics) [to](https://github.com/slackapi/deno-slack-sdk/tree/view-router-generics-overloaded) [do](https://github.com/slackapi/deno-slack-sdk/tree/view-router-uniontype-overloaded) this but all of them failed. This is as close as I've gotten to making this work 😓 😣 😧 

# Changes

- Ordering block kit action type properties alphabetically and adding some docs to the properties
- 🚨 Made Block Kit action payload properties `enterprise` and `is_enterprise_install` NOT optional; they seem to always be present, even in non-enterprise workspaces.
- Renamed the `src/functions/routers` directory to `src/functions/interactivity` for clarity

# TODO

- [x] Types for view payloads and type handlers
- [x] The actual view router code
- [x] 🐛 Userland code trying to pop the exports into place in function code leads to TS complaints; seems to be some issue with the `ViewsRouter()` constructor type assertion magic
- [x] Tests
- [ ] Docs under `docs/`


selfcontained (approved):
Nice work, thanks for doing this.

PR #84, opened 2022-08-22
Simpler functions with handlers - SlackFunctionHandlers

selfcontained:
###  Summary

This provides a new top level export, `SlackFunctionHandlers`, that is intended to unify and simplify how you create your main function handler, along with any additional interactivity handlers in a single export. It builds upon what we currently have, namely:

* The `SlackFunctionHandler<Definition>` type for the main function handler typing
* `BlockActionsRouter()`
* `ViewsRouter()`
* `unhandledEvent()` handler

Examples are probably the easiest way to describe these changes:

For the simplest scenario, you can create your function handler like so:
```ts
import { SlackFunctionHandlers, DefineFunction } from "deno-slack-sdk/mod.ts";

// This can be imported from anywhere as well
export const ApprovalFunction = DefineFunction({....});

export default SlackFunctionHandlers(ApprovalFunction, async ({ inputs }) => {
  // ... main function handler
);
```

For a more complex scenario where your function may have interactive or unhandled event handlers it might look like this:
```ts
export default SlackFunctionHandlers(ApprovalFunction, async ({ inputs }) => {
    // ... main function handler
  },
).addBlockActionsHandler(
  "approve_request", async ({ inputs }) => {
    // ... block_actions handler
  }
).addViewClosedHandler(
  "approval_modal", async ({ view, inputs }) => {
    // ... view_closed handler
  }
).addViewSubmissionHandler(
  "approval_modal", async ({ view, inputs }) => {
    // ... view_submission handler
  }
).addUnhandledEventHandler(({ body }) => {
  console.log("unhandled event handler");
});
```

The return object from calling `SlackFunctionHandlers()` has the following functions available:

* `addBlockActionsHandler()` - matches the `BlockActionsRouter.addHandler()` signature
* `addViewClosedHandler()` - matches the `ViewsRouter.addClosedHandler()` signature
* `addViewSubmissionHandler()` - matches the `ViewsRouter.addSubmissionHandler()` signature
* `addUnhandledEventHandler()` - Takes a callback that is called for any unhandled event. It receives an object that looks like `{ inputs, token, env, body, team_id }` as it's only parameter. I currently have it so it's only allowing for a single handler, so multiple calls will result in the last one winning. Considered a different name, like `setUnhandledEventHandler`, but I tend to prefer the uniformity of `.add*` on these, where you can discover them all via autocomplete. Open to ideas here.

### Testing
To try this out, you can grab this version of the repo and import `SlackFunctionHandlers` to plug in to your function like shown in the examples above. Alternatively to checking this out locally, you can grab the latest from this branch via:

```ts
import { SlackFunctionHandlers } from 'https://raw.githubusercontent.com/slackapi/deno-slack-sdk/bbf412f635bf9681d28c67248cc42ac3a2ab6e5f/src/mod.ts';
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 .coverage
+lcov.info
 npm/
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -14,6 +14,7 @@
   "tasks": {
     // TODO: the bundle command should be `deno bundle src/mod.ts > /dev/null` (less noisy), but redirects are not currently supported in task commands
     "test": "deno fmt --check && deno lint && deno bundle src/mod.ts && deno test src",
-    "coverage": "deno test --coverage=.coverage src && deno coverage --exclude=fixtures --exclude=test .coverage"
+    "coverage": "deno test --coverage=.coverage src && deno coverage --exclude=fixtures --exclude=test .coverage",
+    "coverage-lcov": "deno test --coverage=.coverage src && deno coverage --exclude=fixtures --exclude=test .coverage --lcov --output=lcov.info"
   }
 }
--- a/src/dev_deps.ts
+++ b/src/dev_deps.ts
@@ -2,8 +2,10 @@ export {
   assertEquals,
   assertExists,
   assertMatch,
+  assertRejects,
   assertStrictEquals,
-} from "https://deno.land/std@0.134.0/testing/asserts.ts";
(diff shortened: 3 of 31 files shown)


shapirone (approved):
I opened the can of worms on what to call the thing in the Slack thread, but this LGTM if we decide to keep it as `SlackFunctionHandlers`!

shapirone (approved):

PR #86, opened 2022-08-24
Add support for custom metadata events

shapirone:
###  Summary

This PR adds support for defining events on the App Manifest.

#### Feature Details
When an app uses `chat.postMessage` from inside a function, they can optionally pass a `metadata` object. That `metadata` object will have an `event_type` and an `event_payload`. If the `event_type` matches one of the keys from the app manifest's `events` object, the `event_payload` needs to match the metadata schema.
* If there are required properties that are missing, the message will still successfully post but the metadata will be dropped and the response will include a `warning` and a `response_metadata` object.
  * There are plans to eventually support a "strict mode" that will fail the `postMessage` call entirely, but that's not currently supported.
* If there are additional properties that are not defined in the manifest, the message will be successfully posted with metadata. If validation should catch the extra properties, the Event must set `additionalProperties: false`

#### SDK Usage

```ts
// manifest.ts
export const MyEvent = DefineEvent({
  name: "my_event",
  type: Schema.types.object,
  properties: {
    id: { type: Schema.types.string },
    summary: { type: Schema.types.string },
  },
  required: ["id", "summary"],
});

export default Manifest({
  ...,
  events: [MyEvent], 
});

// function.ts
import { MyEvent } from "../manifest.ts";

...
const post = await client.chat.postMessage({
    channel: inputs.channel,
    text: "Hello this has metadata",
    metadata: {
      event_type: MyEvent, // this turns into the "my_event" string
      event_payload: {
        id: "my_id",
        summary: "This is a summary of the event",
      },
    },
  });
```

### Testing
1. Create an Event on your manifest of object type and verify manifest validation works
2. Turn that object into a custom type and verify manifest validation works
3. Use that event in a `chat.postMessage` call where your structure matches the expectation
4. Change the `chat.postMessage` call where the structure no longer matches
5. BONUS: Set up an event trigger to fire based on the message metadata being posted

### DIFF
--- a/src/events/mod.ts
+++ b/src/events/mod.ts
@@ -1,0 +1,62 @@
+import { SlackManifest } from "../manifest/mod.ts";
+import { ManifestCustomEventSchema } from "../manifest/manifest_schema.ts";
+import {
+  CustomEventDefinition,
+  DefineEventSignature,
+  ICustomEvent,
+} from "./types.ts";
+
+export const DefineEvent: DefineEventSignature = <
+  Def extends CustomEventDefinition,
+>(
+  definition: Def,
+) => {
+  return new CustomEvent(definition);
+};
+
+export class CustomEvent<Def extends CustomEventDefinition>
+  implements ICustomEvent {
+  public id: string;
+  public title: string | undefined;
+  public description: string | undefined;
+
+  constructor(
+    public definition: Def,
+  ) {
+    this.id = definition.name;
+    this.definition = definition;
+    this.description = definition.description;
+    this.title = definition.title;
+  }
+
+  private generateReferenceString() {
+    return this.id;
+  }
+
+  toString() {
+    return this.generateReferenceString();
+  }
+
+  toJSON() {
+    return this.generateReferenceString();
+  }
+
+  registerParameterTypes(manifest: SlackManifest) {
+    if ("properties" in this.definition) {
+      // Loop through the properties and register any types
+      Object.values(this.definition.properties)?.forEach((property) => {
+        if ("type" in property && property.type instanceof Object) {
+          manifest.registerType(property.type);
+        }
+      });
+    } else if (this.definition.type instanceof Object) {
+      // The referenced type is a Custom Type
+      manifest.registerType(this.definition.type);
+    }
+  }
+  export(): ManifestCustomEventSchema {
+    // remove name from the definition we pass to the manifest
+    const { name: _n, ...definition } = this.definition;
+    return definition;
+  }
+}
--- a/src/events/types.ts
+++ b/src/events/types.ts
@@ -1,0 +1,33 @@
+import {
+  CustomTypeParameterDefinition,
+  TypedObjectParameterDefinition,
+} from "../parameters/types.ts";
+import { ManifestCustomEventSchema } from "../manifest/manifest_schema.ts";
+import { CustomEvent } from "./mod.ts";
+import { SlackManifest } from "../manifest/mod.ts";
+
+type AcceptedEventTypes =
+  | TypedObjectParameterDefinition
+  | CustomTypeParameterDefinition;
+
+export type CustomEventDefinition =
+  & {
+    /**
+     * The name of your event
+     * @example my_custom_event
+     */
+    name: string;
+  }
+  & AcceptedEventTypes;
+
+export type DefineEventSignature = {
+  <Def extends CustomEventDefinition>(definition: Def): CustomEvent<Def>;
+};
+
+export interface ICustomEvent {
+  id: string;
+  definition: CustomEventDefinition;
+  description?: string;
+  registerParameterTypes: (manifest: SlackManifest) => void;
+  export(): ManifestCustomEventSchema;
+}
--- a/src/events/types_test.ts
+++ b/src/events/types_test.ts
@@ -1,0 +1,43 @@
+import { DefineEvent } from "./mod.ts";
+import { assertEquals } from "../dev_deps.ts";
+import { DefineType, Schema } from "../mod.ts";
+import { CustomType } from "../types/mod.ts";
+
+Deno.test("DefineEvent accepts object types", () => {
+  const TestEvent = DefineEvent({
+    name: "test",
+    type: Schema.types.object,
+    properties: {},
+  });
+
+  assertEquals(typeof TestEvent.definition.type, "string");
+});
+
+Deno.test("DefineEvent accepts custom types", () => {
+  const TestType = DefineType({
+    name: "test",
+    type: Schema.types.object,
+    properties: {},
+  });
+
+  const TestEvent = DefineEvent({
+    title: "Title",
+    description: "Description",
+    name: "test",
+    type: TestType,
+  });
+
+  assertEquals(TestEvent.definition.type instanceof CustomType, true);
+});
+
+Deno.test("DefineEvent is properly stringified", () => {
+  const TestEvent = DefineEvent({
+    name: "test",
+    type: Schema.types.object,
+    properties: {},
+  });
+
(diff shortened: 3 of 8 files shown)


filmaj:
I was unable to test step 1. Here's my `manifest.ts`:

```typescript
➜ cat manifest.ts
import { Manifest, DefineEvent, Schema } from "deno-slack-sdk/mod.ts";
import { ApprovalWorkflow } from "./workflows/approval.ts";

export const MyEvent = DefineEvent({
  name: "my_event",
  type: Schema.types.object,
  properties: {
    id: { type: Schema.types.string },
    summary: { type: Schema.types.string },
  },
  required: ["id", "summary"],
});

export default Manifest({
  name: "interactive-approval",
  description: "Approving allthethings",
  icon: "assets/icon.png",
  workflows: [ApprovalWorkflow],
  outgoingDomains: [],
  botScopes: ["commands", "chat:write", "chat:write.public"],
  events: [MyEvent],
});
```

And here's the output of the `manifest validate` command (using the latest slack-cli `main` branch):

```
➜ slak manifest validate
Check /Users/fmaj/.slack/slack-debug.log for full error logs

🚫  App manifest generated for your project is invalid. (invalid_manifest)

Error Details:

1: Event Subscription requires a Request URL (requires_request_url)
Source: /settings/event_subscriptions

2: Event Subscription requires Socket Mode if no Request URL is provided (requires_socket_mode_enabled)
Source: /settings/event_subscriptions

```

filmaj:
Looks like there are issues with the `manifest validate` command, should probably update the testing instructions to just use `deploy`.

filmaj:
Testing notes, continued:

Step 2. Using custom type in this way worked fine:

```typescript
const fancyAssBoolean = DefineType({
  name: "fancyAssBoolean",
  type: Schema.types.object,
  properties: {
    aBoolean: { type: "boolean" },
  },
});

export const MyEvent = DefineEvent({
  name: "my_event",
  type: fancyAssBoolean,
  // required: ["aBoolean"],
});
```

If I uncomment the `required` prop in `DefineEvent`, though, I get this error:

```
🚫  App manifest generated for your project is invalid. (invalid_manifest)

Error Details:

1: failed to match exactly one allowed schema (failed_constraint)
Source: /events/my_event
```

Step 3. Use the event in a postMessage metadata. Worked great!

Step 4. Use the event in a postMessage metadata, but where structure doesn't match. So with this Event w/ Custom Type definition:

```typescript
const fancyAssBoolean = DefineType({
  name: "fancyAssBoolean",
  type: Schema.types.object,
  properties: {
    aBoolean: { type: "boolean" },
  },
});

export const MyEvent = DefineEvent({
  name: "my_event",
  type: fancyAssBoolean,
  // required: ["aBoolean"],
});
```

I posted a message like so:

```typescript
  const resp = await client.chat.postMessage({
      channel: inputs.approval_channel_id,
      blocks: renderApprovalMessage(inputs),
      metadata: {
        event_type: MyEvent,
        event_payload: {
          aBoolean: "string"
        },
      }
    });
```

... and no `metadata` is attached to the message at all.

If I posted a message like so:

```typescript
  const resp = await client.chat.postMessage({
      channel: inputs.approval_channel_id,
      blocks: renderApprovalMessage(inputs),
      metadata: {
        event_type: MyEvent,
        event_payload: {
          nope: true
        },
      }
    });
```

Then the above metadata _is_ attached to the message - even though the structure is wrong 🤔

src/events/mod.ts
```diff
@@ -0,0 +1,62 @@
...
+  Def extends CustomEventDefinition,
+>(
+  definition: Def,
+) => {
+  return new CustomEvent(definition);
+};
+
+export class CustomEvent<Def extends CustomEventDefinition>
+  implements ICustomEvent {
+  public id: string;
```
selfcontained, line 19: Should we just call this `name`?

shapirone:
> Step 2. Using custom type in this way worked fine:
> If I uncomment the `required` prop in `DefineEvent`, though, I get this error:

Good find @filmaj, but I consider this unrelated to this PR. It's happening because we don't support overriding properties on the CustomType and is falling back to a generic error. We should share this finding with the primitives team and they can decide what to do about it (improve the error or allow this functionality).

> 
> Step 4. Use the event in a postMessage metadata, but where structure doesn't match. So with this Event w/ Custom Type definition:

> I posted a message like so:
> 
> ```ts
>   const resp = await client.chat.postMessage({
>       channel: inputs.approval_channel_id,
>       blocks: renderApprovalMessage(inputs),
>       metadata: {
>         event_type: MyEvent,
>         event_payload: {
>           aBoolean: "string"
>         },
>       }
>     });
> ```
> 
> ... and no `metadata` is attached to the message at all.

This is expected since `aBoolean: "string"` fails the validation check because it's not a valid `boolean` value. If you `console.log(resp)` I expect you get the warning back!
> 
> If I posted a message like so:
> 
> ```ts
>   const resp = await client.chat.postMessage({
>       channel: inputs.approval_channel_id,
>       blocks: renderApprovalMessage(inputs),
>       metadata: {
>         event_type: MyEvent,
>         event_payload: {
>           nope: true
>         },
>       }
>     });
> ```
> 
> Then the above metadata _is_ attached to the message - even though the structure is wrong 🤔

This is also expected. The `aBoolean` property is not marked as `required`, so validation is okay with it not being passed. Since `fancyAssBoolean ` doesn't specify that `additionalProperties: false` the validation allows unexpected `nope` property to be passed.

src/events/mod.ts
shapirone: I chose `id` because it matches Custom Types and events is pretty much the same. That being said, if we went with `name` it will match Datastores (and the property that's being set). I don't have a strong preference here, open to changing if anyone else does!

src/events/mod.ts
```diff
@@ -0,0 +1,62 @@
...
+  public title: string | undefined;
+  public description: string | undefined;
+
+  constructor(
+    public definition: Def,
+  ) {
+    this.id = definition.name;
```
selfcontained, line 26: Should we just call this name?

src/events/mod.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  toString() {
+    return this.generateReferenceString();
+  }
+
+  toJSON() {
+    return this.generateReferenceString();
+  }
+
```
jmplexi, line 43: Is it intended for toString() and toJSON() to do the same thing?

src/events/mod.ts
shapirone: I chose `id` because it matches Custom Types and events is pretty much the same. That being said, if we went with `name` it will match Datastores (and the property that's being set). I don't have a strong preference here, open to changing if anyone else does!

src/events/mod.ts
shapirone: Yep! Overwriting these prototype methods is a pattern we follow elsewhere (like in types and workflows) to make sure that when a developer references the `CustomEvent` instance, we convert it to what our API endpoints expect for the manifest and API clients rather than the full object.

`toJSON()` supports the case where we convert an object into a string to pass to our APIs
`toString()`supports the case where devs may want to do something like `This is the event: ${MyEvent}`

filmaj (approved):
The testing steps, including creating a custom event w/ custom type, and registering an event trigger on that thing, worked. Well done sir.

src/events/mod.ts
selfcontained: Consistency is good, so am fine w/ this, just couldn't recall if we had odd naming in Types due to callback_id => id => name refactoring, so thought if so, maybe could keep this clearer here.

selfcontained (approved):
Looks good, and works as described.

PR #87, opened 2022-08-24
Check for additionalProperties to be set to false

shapirone:
###  Summary

The backend defaults `additionalProperties` on parameters of the `object` type to true, so this PR inverts the check to see if they are explicitly setting `additionalProperties` to false.

### Testing
1. Create an `object` parameter as a function input or output
2. Verify that at both workflow authoring time and at runtime, you're able to access unexpected properties.
3. Change the object parameters you created to set `additionalProperties: false`
4. Notice that your access of unexpected properties now results in errors.

```ts
anObject: {
        type: Schema.types.object,
        properties: {
            test: { type: Schema.types.string },
        },
        additionalProperties: false,
      },
```

### DIFF
--- a/src/functions/interactivity/block_actions_types.ts
+++ b/src/functions/interactivity/block_actions_types.ts
@@ -76,6 +76,7 @@ export type BlockActionsBody = {
     id: string;
     /**
      * @description User's handle as seen in the Slack client when e.g. at-notifying the user.
+     */
     name: string;
     /**
      * @description The encoded team ID for the workspace, or team, where the Block Kit action originated from.
--- a/src/functions/slack_function_handler_type_test.ts
+++ b/src/functions/slack_function_handler_type_test.ts
@@ -1,4 +1,4 @@
-import { assertExists } from "../dev_deps.ts";
+import { assertEquals, assertExists } from "../dev_deps.ts";
 import { assertEqualsTypedValues } from "../test_utils.ts";
 import { SlackFunctionTester } from "./tester/mod.ts";
 import { DefineFunction } from "./mod.ts";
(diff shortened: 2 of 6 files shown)


src/functions/interactivity/block_actions_types.ts
```diff
@@ -76,6 +76,7 @@ export type BlockActionsBody = {
     id: string;
     /**
      * @description User's handle as seen in the Slack client when e.g. at-notifying the user.
+     */
```
shapirone, line 79: There was an issue here where we weren't closing the comment, which meant that `name` wasn't part of the BlockActionsBody defined payload. This is unrelated to the core of this PR and can be removed if we want.

src/functions/slack_function_handler_type_test.ts
```diff
@@ -371,3 +371,125 @@ Deno.test("SlackFunctionHandler using Custom Types", () => {
...
+  };
+
+  const { createContext } = SlackFunctionTester(TestFunction);
+
+  const result = handler(createContext({ inputs: sharedInputs }));
+  assertEqualsTypedValues(sharedInputs, result.outputs);
+  assertExists(result.outputs?.noAddlPropertiesObj);
+  assertExists(result.outputs?.noAddlPropertiesObj.aString);
+
+  // @ts-expect-error anythingElse cant exist
```
shapirone, line 493: I'm not sure there's a better way to do this type of test, but if this directive becomes unused (i.e. no error), it will let us know!

src/functions/types.ts
```diff
@@ -93,15 +93,20 @@ type UnknownRuntimeType = any;
...
   }
-  & {
-    [key: string]: UnknownRuntimeType;
-  };
+  : 
+    & {
+      [k in keyof Param["properties"]]: FunctionInputRuntimeType<
+        Param["properties"][k]
+      >;
+    }
```
shapirone, line 106: We could probably figure out some way to avoid this duplication, but I didn't want to spend too much time on this. Open to suggestions.

src/functions/slack_function_handler_type_test.ts
selfcontained: TIL

selfcontained (approved):
beautiful, works as described.

PR #88, opened 2022-08-25
Make the workflow callback_id generic

shapirone:
###  Summary

This makes it so that we respect the value passed into `callback_id`, rather than treating it as any old `string`. This is necessary for us to key off of `callback_id` to know what should be passed into a trigger's `workflow` parameter in the API.

#### Testing
There is no functional difference here, but you could test making and deploying workflows for a sanity check.


filmaj (approved):

selfcontained (approved):

jmplexi:
LGTM

jmplexi (approved):
LGTM

PR #90, opened 2022-08-29
Default token_management_enabled:true for runOnSlack:false apps

srajiang:
###  Summary

Removes `tokenManagementEnabled` top level property from the Manifest definition, and sets corresponding token_management_enabled property in the Manifest Schema as true for runOnSlack: false apps. 

Samples with `tokenManagementEnabled` field should also be updated. 

### DIFF
--- a/src/manifest/manifest_test.ts
+++ b/src/manifest/manifest_test.ts
@@ -246,6 +246,21 @@ Deno.test("Manifest() always sets token_management_enabled to false for function
   assertEquals(manifest.oauth_config.token_management_enabled, false);
 });
 
+Deno.test("Manifest() sets token_management_enabled to true by default for runOnSlack: false apps", () => {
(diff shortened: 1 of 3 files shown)


stevengill:
Could remote apps one day decide they want to disable token management? Allowing slack to manage it for them? I would assume this "remote app with token management disabled" would only rely on JIT tokens, no bot token

selfcontained (approved):
Thanks!

srajiang:
@stevengill - Not out of the realm of possibility, but it seems sufficiently far off that we can wait to see. The door isn't closed to re-introducing the `tokenManagementEnabled` flag after this change, but if we roll into beta letting Bolt devs set tokenManagementEnabled to false, I don't see how they could deploy their apps today, since no `xoxb` will be visible to them.

stevengill (approved):

PR #91, opened 2022-08-29
Remove some deprecated items

selfcontained:
###  Summary

This removes 2 deprecated items:

* Top level `FunctionHandler` export that was replaced with `SlackFunctionHandler` type.
* CustomType definitions support for `callback_id` that was replaced with `name`.

### DIFF
--- a/src/types/mod.ts
+++ b/src/types/mod.ts
@@ -23,7 +23,7 @@ export class CustomType<Def extends CustomTypeDefinition>
   constructor(
     public definition: Def,
   ) {
-    this.id = "name" in definition ? definition.name : definition.callback_id;
+    this.id = definition.name;
     this.definition = definition;
     this.description = definition.description;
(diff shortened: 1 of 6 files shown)


src/types/mod.ts
```diff
@@ -23,7 +23,7 @@ export class CustomType<Def extends CustomTypeDefinition>
   constructor(
     public definition: Def,
   ) {
-    this.id = "name" in definition ? definition.name : definition.callback_id;
+    this.id = definition.name;
```
shapirone, line 26: We should also remove `callback_id` support from the `export()` method

shapirone (changes requested):
Let's remove handling from the Custom Type `export` method as well. Should we do a doc sweep looking for `callback_id` or `FunctionHandler`?

shapirone (approved):
LGTM!

PR #93, opened 2022-08-30
Remove home_tab_enabled for runOnSlack apps

srajiang:
###  Summary

Ensures `home_tab_enabled` doesn't show up during RunOnSlack app manifest authoring as the related app_home_opened event isn't supported at the moment for ROSI apps. 

### DIFF
--- a/src/manifest/mod.ts
+++ b/src/manifest/mod.ts
@@ -279,11 +279,9 @@ export class SlackManifest {
         messagesTabReadOnlyEnabled,
       } = def.features.appHome;
 
-      if (manifest.features.app_home !== undefined) {
-        manifest.features.app_home.messages_tab_enabled = messagesTabEnabled;
-        manifest.features.app_home.messages_tab_read_only_enabled =
-          messagesTabReadOnlyEnabled;
-      }
+      manifest.features.app_home.messages_tab_enabled = messagesTabEnabled;
+      manifest.features.app_home.messages_tab_read_only_enabled =
+        messagesTabReadOnlyEnabled;
     }
 
     // External Auth providers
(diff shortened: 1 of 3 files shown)


src/manifest/mod.ts
```diff
@@ -52,10 +54,6 @@ export class SlackManifest {
         bot_user: {
           display_name: def.displayName || def.name,
         },
-        app_home: {
```
srajiang, line 55: This auto-defaults remote apps to these values as well unless explicitly set by the dev which seems unintentional. I have moved the logic setting these defaults to to ROSI specific `assignRunOnSlackManifestProperties`.

src/manifest/mod.ts
```diff
@@ -111,21 +109,6 @@ export class SlackManifest {
       );
     }
 
-    if (def.features?.appHome) {
```
srajiang, line 114: We should probably not be setting these overrides all in one place since it's not clear that there's a distinction for ROSI vs. Remote apps. Prefer to handling anything ROSI specific in the `assignRunOnSlackManifestProperties `and Remote specific in `assignRemoteSlackManifestProperties`

srajiang (commented):
Left some comments as well for reviewers!

src/manifest/mod.ts
```diff
@@ -264,6 +262,30 @@ export class SlackManifest {
...
+    } as ManifestAppHomeMessagesTabSchema;
+
+    // Allow App Home override values if provided in apphome
+    if (def.features?.appHome) {
+      const {
+        messagesTabEnabled,
+        messagesTabReadOnlyEnabled,
+      } = def.features.appHome;
+
+      if (manifest.features.app_home !== undefined) {
```
selfcontained, line 282: With the defaulting now happening right above this, we can remove this if statement since it will never be undefined

selfcontained (approved):
Looks good - one non-blocking comment.

shapirone (approved):
LGTM, no longer seeing `homeTabEnabled` and the validation around message tab is still working!

src/manifest/mod.ts
srajiang: Yup! Removed it.

PR #95, opened 2022-08-31
Adds a document outlining Events

shapirone:
###  Summary

Introduces a document describing how to use Events


docs/events.md
```diff
@@ -0,0 +1,132 @@
+## Events
+
+Custom Events provide Schema validation for an App's [message metadata](https://api.slack.com/metadata).
+
+### Defining a Event
+
+Events can be defined with the top level `DefineEvent` export. Events must be set up as an `object `type or a [`custom Type`][types] of an `object` type. Below is an example of setting up a custom Event that can be used during an incident.
```
shapirone, line 7:
```suggestion
Events can be defined with the top level `DefineEvent` export. Events must be set up as an `object` type or a [`custom Type`][types] of an `object` type. Below is an example of setting up a custom Event that can be used during an incident.
```

docs/events.md
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+});
+```
+
+```ts
+// At function runtime
+// This example assumes all required values are passed to the function's inputs
+client.chat.postMessage({
+  {
```
filmaj, line 74: I think one pair of brackets here are extra?

filmaj (changes requested):
Nice work! Left a few comments.

docs/events.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+### Referencing Events
+
+There are two places where you can reference your events:
+1. Posting a message to Slack
+2. Creating a Message Metadata Trigger
+
+#### Posting a message to Slack
+Event validation happens against the App's manifest when an App posts a message to Slack using the `metadata` parameter. If the `event_type` matches a custom Event specified in the App's manifest, it will validate that all required parameters are provided. If it doesn't meet the validation standards, a warning will be returned in the response and the metadata will be dropped from the message.
```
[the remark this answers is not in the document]
ashleymcm: Not sure where this fits in, but metadata can also be sent/changed when updating a message. The behaviour is otherwise the same as when posting a message.

docs/events.md
```diff
@@ -0,0 +1,123 @@
...
+  inputs: {
+    id: "{{data.metadata.event_payload.incident_id}}",
+    title: "{{data.metadata.event_payload.incident_title}}",
+    summary: "{{data.metadata.event_payload.incident_summary}}",
+    severity: "{{data.metadata.event_payload.incident_severity}}",
+    date_created: "{{data.metadata.event_payload.incident_date}}"
+  },
+  workflow: "#/workflows/start_incident",
+  event: {
+    event_type: "slack#/events/message_metadata_posted",
```
ashleymcm, line 110: Again just a small note - we can listen here for posted, updated, or deleted. Not sure where is best to add that detail 😅

ashleymcm (commented):
This is really nice documentation, thank you 🙌

docs/events.md
betsyalegria: ^ that's true for the Events API but for triggers, only `message_metadata_posted` is supported right now 😅

shapirone: Yeah, planning to add updated and deleted when those are supported as triggers 🙇

docs/events.md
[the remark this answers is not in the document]
shapirone: @filmaj I made these updates — it's done against the `name` property

filmaj (approved):

jesselawson:
Would it be more accurate to call these `Message metadata types` rather than events? That they're called events makes it seem like we'd define them in an `events` folder, but based on what they are, I wonder if it would be more appropriate to put them in a `types` folder?

jesselawson:
Also posted a Q [here](https://slack-pde.slack.com/archives/C03RA1T9DK5/p1662501246529529) about using Custom types: 

Trying to get Custom types and `DefineEvent` to work correctly, but it seems like I'm missing something:
- I created a custom type `IncidentType`
- If I create an event with `DefineEvent` that has type `Schema.types.object` with properties identical to `IncidentType` , then `hermes run` works as expected
- If I change the type to `IncidentType`, despite all properties being identical, I receive the following error when trying to execute `hermes run`: `failed to match exactly one allowed schema\nSource: /events/incident`

selfcontained:
> Would it be more accurate to call these `Message metadata types` rather than events? That they're called events makes it seem like we'd define them in an `events` folder, but based on what they are, I wonder if it would be more appropriate to put them in a `types` folder?

I think the longer term vision is that they do become more of an event primitive of the platform, and have some discoverability around them as well. This led to them being defined under the `events` property on the app manifest. I think there's value in keeping the naming consistent with how we've chosen to identify them as part of an app's manifest. I agree it isn't obvious how they connect to message metadata "event_type" matching in the current state of things though, but longer term I hope that becomes more clear or less necessary for someone to know about.

PR #96, opened 2022-09-01
Fix parameter typing based on the discriminate union

shapirone:
###  Summary

Now when you set a particular `type`, you'll get the specific properties for that `type` back in typeahead

### Testing

When defining a parameter (function input/output, event, type, etc.), you should now see the relevant typeahead
image


selfcontained (approved):
Looks good, thanks!

filmaj (approved):
Beauty, works for me nicely! AFAICT the only tweaks here are for strings and numbers, yes? I see min/maxLength now available exclusively for strings, and minimum/maximum exclusively for numbers.

selfcontained:
> Beauty, works for me nicely! AFAICT the only tweaks here are for strings and numbers, yes? I see min/maxLength now available exclusively for strings, and minimum/maximum exclusively for numbers.

That's mostly it - there's a few types that have `enum` and `choices`, and array has a few custom props as well. It should match [these type-specific type defs](https://github.com/slackapi/deno-slack-sdk/blob/56e0b848871bc227093cb406be5eb5cf6906f741/src/parameters/types.ts#L68-L134) now

PR #98, opened 2022-09-01
Revert "Fix parameter typing based on the discriminate union (#96)"

shapirone:
This reverts commit 895e3d9484eeb720a6e6d118bc13960b6649b332.

###  Summary
Reverts this "fix"

### Testing
This commit introduced a bug where two workflows with parameters being used will result in an infinite loop error. Verify that using two workflows with parameters will not lead to an error.

Note that this reintroduces the issue where Parameters aren't properly typed based on the value provided to the `type` property.


selfcontained (approved):
LGTM - tested and it does remove the type error introduced.

PR #103, opened 2022-09-13
Try 2: Fix parameter typing based on the discriminate union

shapirone:
This reapplies the change from #96 but resolves the bug where multiple workflows couldn't be defined.

###  Summary
Now when you set a particular `type`, you'll get the specific properties for that `type` back in typeahead

### Testing

When defining a parameter (function input/output, event, type, etc.), you should now see the relevant typeahead
image


filmaj (approved):
:shipit:

selfcontained (approved):
Looks good - verified it doesn't have that error w/ multiple workflows, and we have a test for that now anyways.

PR #105, opened 2022-09-15
Add specific instructions on how to implement a function in docs

filmaj:
Also add , add team_id to list of function context variables.


jesselawson (approved):
Thanks for updating this!

PR #106, opened 2022-09-15
[HERMES-3842] Add npm build action

srajiang:
###  Summary

Adds a git action that runs the build_npm.ts script on PRs and commits to `main`. 

The Npm Build step this adds should fail currently, this is expected. A [separate PR](https://github.com/slackapi/deno-slack-sdk/pull/107) will fix the issues causing the action failure.

### Requirements (place an `x` in each `[ ]`)

* [x] I've read and understood the [Contributing Guidelines](https://github.com/{project_slug}/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #107, opened 2022-09-16
[HERMES-3843] Shim deno fetch api for npm build

srajiang:
###  Summary

@fil discovered that - `build_npm.ts` which uses `dnt` package to build deno modules for npm distribution - was failing due to some recent changes that added deno-slack-api as an overall project dependency.

This PR resolves that issue by adding a polyfill for that dependency. 

Related to [this PR](https://github.com/slackapi/deno-slack-sdk/pull/106) which adds an extra check on PRs to main to run the build. 

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -1,0 +1,31 @@
+# This workflow runs a test build for npm against changes on main or PRs
+
+name: Npm Build
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: macos-latest
+    
+    steps:
+      - name: Actions checkout
+        uses: actions/checkout@v3
+      
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: latest
+          registry-url: https://registry.npmjs.org/
+          
+      - name: Setup Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: v1.x
+
+      - name: Run build_npm.ts
+        run: deno run -A build_npm.ts
--- a/build_npm.ts
+++ b/build_npm.ts
@@ -7,6 +7,10 @@ await build({
   typeCheck: false,
   entryPoints: ["./src/mod.ts"],
   outDir: "./npm",
+  // ensures that the emitted package is compatible with node v14 later
+  compilerOptions: {
+    target: "ES2020"
+  },
   shims: {
     // see JS docs for overview and more options
     deno: true,
--- a/build_npm.ts
+++ b/build_npm.ts
@@ -13,9 +13,9 @@ await build({
     // custom shims
     custom: [{
       package: {
-        name: "node-fetch-commonjs",
+        name: "node-fetch",
         // Please see: https://www.npmjs.com/package/node-fetch-commonjs
-        version: "~3.1.1", 
+        version: "2.6.7", 
       },
       globalNames: [{
         name: "fetch",
--- a/build_npm.ts
+++ b/build_npm.ts
@@ -4,11 +4,24 @@ import { build, emptyDir } from "https://deno.land/x/dnt/mod.ts";
 await emptyDir("./npm");
 
 await build({
+  typeCheck: false,
   entryPoints: ["./src/mod.ts"],
   outDir: "./npm",
   shims: {
     // see JS docs for overview and more options
     deno: true,
+    // custom shims
+    custom: [{
+      package: {
+        name: "node-fetch-commonjs",
+        // Please see: https://www.npmjs.com/package/node-fetch-commonjs
+        version: "~3.1.1", 
+      },
+      globalNames: [{
+        name: "fetch",
+        exportName: "default"
+      }]
+    }],
   },
   package: {
     // package.json properties


filmaj:
Hmm, I'm having issues running this locally:

```
➜ deno run -A build_npm.ts 3.3.3
[dnt] Transforming...
[dnt] Running npm install...
npm notice created a lockfile as package-lock.json. You should commit this file.
added 16 packages from 54 contributors and audited 16 packages in 0.705s

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

[dnt] Building project...
[dnt] Emitting declaration files...
[dnt] Emitting ESM package...
[dnt] Emitting script package...
[dnt] Running tests...

> @slack/deno-slack-sdk@3.3.3 test /Users/fmaj/src/deno-slack-sdk/npm
> node test_runner.js

Running tests in ./script/datastore/datastore_test.js...

/Users/fmaj/src/deno-slack-sdk/npm/script/deps/deno.land/std@0.152.0/testing/mock.js:124
    id ??= (sessions.length || 1) - 1;
       ^^^

SyntaxError: Unexpected token '??='
    at wrapSafe (internal/modules/cjs/loader.js:1001:16)
    at Module._compile (internal/modules/cjs/loader.js:1049:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/Users/fmaj/src/deno-slack-sdk/npm/script/dev_deps.js:33:29)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @slack/deno-slack-sdk@3.3.3 test: `node test_runner.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @slack/deno-slack-sdk@3.3.3 test script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/fmaj/.npm/_logs/2022-09-16T12_30_45_616Z-debug.log
error: Uncaught (in promise) Error: npm run test failed with exit code 1
        throw new Error(
              ^
    at runCommand (https://deno.land/x/dnt@0.28.0/lib/utils.ts:56:15)
    at async build (https://deno.land/x/dnt@0.28.0/mod.ts:364:5)
    at async file:///Users/fmaj/src/deno-slack-sdk/build_npm.ts:6:1
```

srajiang:
@filmaj - Hmm, yeah I'm still having issues repro-ing the error you're seeing locally. I just cherry-picked the changes of the other PR onto this one to see whether the CI/CD with the build step added passes, and it seems to be[ running fine](https://github.com/slackapi/deno-slack-sdk/actions/runs/3070953996/jobs/4961197664).

filmaj:
It was the node version. I was running node 14, and looks like the [logical nullish assignment operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment) (which is what my local run was complaining about, the weird looking `??=` operator) is supported in node 15+.

As long as we're cool with that requirement, let's go ahead. Otherwise, might have to tweak this. If the idea is to integrate with bolt-js or any of our other node SDKs, then we probably want to set a compilation target that supports the minimum node versions that we support there.

srajiang:
@filmaj - Thanks for following up on this - I am catching back up on this thread. I don't have a strong instinct on the right thing to do here, so I'm curious what you think given that we're aim to remove support for v12 imminently, v14 reaches EOL fairly soon as well after the holidays (4/2023), and maybe also given the types consolidation work you've been noodling on?

filmaj:
@srajiang I found a way to make this work on node 14. Let's customize the compiler target by adding an option like this:

```
  ...
  },
  compilerOptions: {
    target: "ES2020",
  },
  package: {
  ...
```

By default, [`dnt` uses ES2021 as the target if unspecified](https://deno.land/x/dnt@0.31.0/mod.ts?source#L158). Played around with this option a bit, here is what I found:
- if we set the compiler target to "ES2020", it works on node v14.20.. but not on node v12.22.12.
- I tried using "ES2019" but at that point some core deno-shim package errors out with 'module not found' when it tries to shim in a compatibility layer for node's `fs` module.  Even when adding the `esModule: false` option to the compiler, still doesn't fix the issue. I'm not sure but I think that tells me that deno's `dnt` module, and specifically its node compatibility layer, has some minimal ECMAscript requirements that we probably can't get around.

.github/workflows/npm.yml
```diff
@@ -0,0 +1,31 @@
...
+          node-version: latest
+          registry-url: https://registry.npmjs.org/
+          
+      - name: Setup Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: v1.x
+
+      - name: Run build_npm.ts
+        run: deno run -A build_npm.ts
```
filmaj, line 31: Running this without another argument, the `Deno.args[0]` we pass in the script to the `version` field becomes undefined and isn't emitted in the package.json. Not sure if that's relevant.

build_npm.ts
```diff
@@ -4,11 +4,24 @@ import { build, emptyDir } from "https://deno.land/x/dnt/mod.ts";
...
+        name: "node-fetch",
+        // Please see: https://www.npmjs.com/package/node-fetch-commonjs
+        version: "2.6.7", 
+      },
+      globalNames: [{
+        name: "fetch",
+        exportName: "default"
+      }]
+    }],
   },
```
filmaj, line 25: Let's add an explicit compilation target here, like 
```suggestion
  },
    compilerOptions: {
    target: "ES2020",
  },
```

This makes the emitted package work on node v14 (though it won't work on node v12).
In combination with the above change, can we also bump up the minimum `engines` node version in `package.json` to match the minimum node version that supports ES2020 (not sure exactly which version that is - at least node v14, but may be even a node v13 something, though I wouldn't spend too much time trying to figure that out, I doubt any sane person is using the odd, unstable node versions).

filmaj (approved):
Left a few comments/suggestions to get at least compatibility for node v14! If those can be added, then this look great to me!

.github/workflows/npm.yml
srajiang: Yeah true - I figured since the purpose of this workflow step is just to ensure that the build completes, we don't have to populate the version.

srajiang (commented):
Thanks so much for helping to figure out how to make this compatible with node v14! I've made the changes, test are passing so it looks like we're ready to go :D 

I'll go ahead and bump the engines fields within bolt-js accordingly.

PR #108, opened 2022-09-28
Adds min supported engines properties to npm build package.json

srajiang:
###  Summary

Adds min supported engines properties to npm build package.json

### DIFF
--- a/build_npm.ts
+++ b/build_npm.ts
@@ -40,6 +40,11 @@ await build({
     bugs: {
       url: "https://github.com/slackapi/deno-slack-sdk/issues",
     },
+    // sets the minimum engine to node v14
+    engines: {
+      "node": ">=14.19.0",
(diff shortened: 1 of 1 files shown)


build_npm.ts
```diff
@@ -40,6 +40,11 @@ await build({
     bugs: {
       url: "https://github.com/slackapi/deno-slack-sdk/issues",
     },
+    // sets the minimum engine to node v14
```
filmaj, line 43: I'd add a note in this comment as to _why_ we set this minimum (because dnt transpilation-generated code seems to only be able to successfully compile as far back ES2020)

filmaj (approved):
LGTM! Left a little nit.

build_npm.ts
srajiang: Thanks! I've added it :)

PR #109, opened 2022-09-29
Add FunctionSourceFile as an easy way to configure source_file in DefineFunction

seratch:
###  Summary

This pull request adds a new utility that enables developers to easily configure the "source_file" argument in `DefineFunction`. Specifically, you can use it in the following way. 

```typescript
import { DefineFunction, FunctionSourceFile, Schema } from "deno-slack-sdk/mod.ts";

const def = DefineFunction({
  callback_id: "hello-world",
  title: "Hello world function",
  // This resolves "functions/hello_world.ts" for you
  // When you rename the file, the value can be automatically changed
  source_file: FunctionSourceFile(import.meta.url),
  input_parameters: { properties: { channel: { type: Schema.slack.types.channel_id } }, required: [] },
  output_parameters: { properties: {}, required: [] },
})
```

As mentioned in the comment in the above code, the benefit of this is that you don't need to update the string when you rename the function source file name. Let's say a developer started with `hello_world.ts` in the beginning, and then the dev renamed it to a more realistic one like `create_channel.ts`. In this case, updating the source_file part in code is necessary too. Contrarily, if the dev goes with this `FunctionSourceFile`, the dev no longer needs to modify any parts of the code just because of renaming a file.

Discussion points in reviews:
* What do we think about the name `FunctionSourceFile`? Any good alternative?
* Can the SDK automatically resolve the file path and allow developers to skip the argument? Perhaps, it's not feasible. IIUC, there is no way to get the file path of a method caller on the callee side.

### DIFF
--- a/src/functions/function_source_file.ts
+++ b/src/functions/function_source_file.ts
@@ -1,0 +1,20 @@
+/**
+ * A factory function for easily generating DefineFunction's source_file parameter value.
+ *
+ * @param importMetaUrl the value of import.meta.url
+ * @param depth pass the depth when having nested directories under functions directory (default: 0)
+ * @returns a valid string value for DefineFunction's source_file argument
+ */
+const FunctionSourceFile = function (
+  // Pass the value of import.meta.url in a function code
+  importMetaUrl: string,
+  // If you have sub diretories under "functions" dir, set the depth.
+  // When you place functions/pto/data_submission.ts, the depth for the source file is 1.
+  depth = 0,
+): string {
+  const sliceStart = -2 - depth;
+  const path = new URL("", importMetaUrl).pathname;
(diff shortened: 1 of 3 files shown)


seratch:
The Deno CI build failure seems to be due to some breaking changes in Deno 1.26. I've confirmed that `deno test` succeeds with Deno 1.25. Perhaps, we should resolve the issues with Deno 1.26 in a separate PR. 

I will look into the npm build issue.

seratch:
I've resolved the test failures with npm builds. The Deno failures will be resolved once https://github.com/slackapi/deno-slack-sdk/pull/110 is merged and this PR is rebased on the latest revision after that.

gmrhom19 (approved):

gmrhom19 (approved):

PR #111, opened 2022-10-03
Expose enterprise_id to userland handlers.

filmaj:
Relies on the work in https://github.com/slackapi/deno-slack-runtime/pull/38.

Exposes an `enterprise_id` property to all handlers.


selfcontained (approved):
Whoa nice, it's just tests, types, and the test helper.

filmaj:
Ya, the runtime does the harder work of parsing through parsing, understanding when different fields exist in what payload, etc. That separation is kinda nice, as we can focus on exposing nicely documented and structured types for devs in this repo only.

shapirone (approved):

filmaj:
The runtime changes are now in production, merging!

PR #112, opened 2022-10-06
Updating interactivity docs for block actions and view events to use the SlackFunction construct

filmaj:
Also: handler-registration methods that chain off the SlackFunction instead of documenting the old Router constructs.

I will send a similar PR to webapp to update the docs on api.slack.com/future as well shortly.


shapirone (approved):
Awesome, LGTM! Thanks for updating this 🙇

PR #113, opened 2022-10-11
Added message_ts and message_context

cchensh:
###  Summary

Added `message_ts` to `SlackPrimitiveTypes` and `message_context` to `CustomSlackTypes`. 

Created new `MessageContextType` as the new Custom Type.

Requirements can be found below: 

- `message_ts` should be added to SlackPrimitiveTypes as a simple string
- `message_context` should be added to CustomSlackTypes as a Custom Type definition of an object with properties that can be found here:
https://corp.quip.com/9T6nAE8mUHxn/Reply-In-Thread-Builtin-Function#temp:C:VKKb19fca4ebf8c44c2b9b6732bc

### DIFF
--- a/src/schema/slack/types/custom/message_context.ts
+++ b/src/schema/slack/types/custom/message_context.ts
@@ -16,6 +16,7 @@ const MessageContextType = DefineType({
       type: SlackPrimitiveTypes.channel_id,
     },
   },
+  required: ["message_ts"],
 });
 
 export { MessageContextType };
--- a/src/schema/slack/types/custom/message_context.ts
+++ b/src/schema/slack/types/custom/message_context.ts
@@ -1,0 +1,21 @@
+import SchemaTypes from "../../../schema_types.ts";
+import { SlackPrimitiveTypes } from "../../types/mod.ts";
+import { DefineType } from "../../../../types/mod.ts";
+
+const MessageContextType = DefineType({
+  name: "slack#/types/message_context",
+  type: SchemaTypes.object,
+  properties: {
+    message_ts: {
+      type: SlackPrimitiveTypes.message_ts,
+    },
+    user_id: {
+      type: SlackPrimitiveTypes.user_id,
+    },
+    channel_id: {
+      type: SlackPrimitiveTypes.channel_id,
+    },
+  },
+});
+
+export { MessageContextType };
--- a/src/schema/slack/types/mod.ts
+++ b/src/schema/slack/types/mod.ts
@@ -7,7 +7,7 @@ const SlackPrimitiveTypes = {
   blocks: "slack#/types/blocks",
   oauth2: "slack#/types/credential/oauth2",
   rich_text: "slack#/types/rich_text",
-  message_ts: "string",
+  message_ts: "slack#/types/message_ts",
 } as const;
 
 export { SlackPrimitiveTypes };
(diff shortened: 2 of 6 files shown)


src/schema/slack/types/mod.ts
```diff
@@ -7,6 +7,7 @@ const SlackPrimitiveTypes = {
   blocks: "slack#/types/blocks",
   oauth2: "slack#/types/credential/oauth2",
   rich_text: "slack#/types/rich_text",
+  message_ts: "string",
```
shapirone, line 10: This needs to be `message_ts: "slack#/types/message_ts"`

src/schema/slack/types/custom/message_context.ts
```diff
@@ -0,0 +1,21 @@
...
+    message_ts: {
+      type: SlackPrimitiveTypes.message_ts,
+    },
+    user_id: {
+      type: SlackPrimitiveTypes.user_id,
+    },
+    channel_id: {
+      type: SlackPrimitiveTypes.channel_id,
+    },
+  },
```
shapirone, line 18: We should add `required: ["message_ts"]` here since that always must be provided. The others are optional

src/schema/slack/types/mod.ts
cchensh, line 10: From the doc it looks like we should use `slack#/types/message_ts` type instead of plain string? @shapirone?

shapirone (changes requested):
Sorry, forgot to mention one other spot — `FunctionInputRuntimeType` needs to know to treat `message_ts` as a `string` at runtime. You can add a check for `message_ts` at [this line](https://github.com/slackapi/deno-slack-sdk/blob/main/src/functions/types.ts#L91).

When these updates are made this will be looking pretty good and we just need to verify it works e2e in the client.

filmaj:
Hello after a few months @cchensh (cc @shapirone)!

So we just merged in the big SDK refactor we worked on. I rebased this branch on the latest main and pushed it up to your branch (hope you don't mind). Thanks to Neil, he gave me an idea on how to test this PR:

1. Pull in the latest `functions.list` API payload. That has a reply-in-thread built-in function definition which I used to generate the latest built-in functions available.
2. In the ensuing `git diff`, it looks like the SendMessage built-in now provides a message-context output parameter! 👀 
3. There's a new reply-in-thread built-in function that takes as input a message-context parameter.

So, I made a small change to the hello world / send a greeting sample so that the bottom of its workflow definition looks like this:

```
const msgStep = GreetingWorkflow.addStep(Schema.slack.functions.SendMessage, {
  channel_id: inputForm.outputs.fields.channel,
  message: greetingFunctionStep.outputs.greeting,
});

GreetingWorkflow.addStep(Schema.slack.functions.ReplyInThread, {
  message_context: msgStep.outputs.message_context,
  message: "hey look i can reply in thread using message_context now too woot",
});

export default GreetingWorkflow;
```

.. and that seems to work! So I think this PR is functionally OK 👍 

Screenshot 2023-02-02 at 4 17 06 PM

I'm gonna look at adding a few more tests to this PR (in our fancy new `tests/integration` directory, to make sure the sub-properties on message-context are of the correct type and whatnot) and then I think it is good to merge.

filmaj (approved):

PR #116, opened 2022-10-20
Add support for block suggestion handlers

filmaj:
###  Summary

This PR adds a new method available to app functions: `addBlockSuggestionHandler`.

It is a copy-paste of the existing Block Actions handler support, basically.

After merging this, we will have in place all of the available interactivity handlers. Thus, after this gets merged would be a good opportunity to review all the interactivity code and see how we should refactor it.

## Testing

1. Pull down this branch, and point a hermes app's `import_map.json` to the location of this checked out branch + `/src/`, e.g. `file:///Users/fmaj/src/deno-slack-sdk/src/`
2. Add an external data source select menu to a message or view your app creates. Perhaps it can post a message that includes this in its Block Kit payload:

```ts
      {
        type: "input",
        block_id: "ext_select_block",
        optional: true,
        element: {
          type: "external_select",
          action_id: "ext_select_input",
          placeholder: {
            type: "plain_text",
            text: "Inspire",
          },
        },
        label: {
          type: "plain_text",
          text: "Inspirational Quote",
        },
      },
```

3. Register a block suggestion handler in your app to respond to interactions with the above drop down menu. The following example keys off of the above block kit snippet's `action_id`:

```ts
export default SlackFunction(...).addBlockSuggestionHandler(
  "ext_select_input", // <-- the action id of the select drop down menu
  async (args) => { // <-- `args` here should be nicely typed, and block_suggestion-specific payload bits available under `args.body`
    console.log('BLOCK SUGGESTION HANDLER, ', JSON.stringify(args.body, null, 2));
    // Fetch an inspirational quote
    const apiResp = await fetch("https://motivational-quote-api.herokuapp.com/quotes");
    const quotes = await apiResp.json();
    console.log('Returning', quotes.length, 'quotes');
    const opts = {
      "options": quotes.map((q) => ({value: `${q.id}`, text: {type:"plain_text", text: q.quote.slice(0,70)}})) // <-- watch out! options have a max text length of 75 characters
    };
    return opts;
  }
);
```


docs/functions-suggestion-handlers.md
```diff
@@ -0,0 +1,218 @@
...
+}).addBlockSuggestionHandler(
+  "ext_select_input", // The first argument to addBlockActionsHandler can accept an action_id string, among many other formats!
+  // Check the API reference at the end of this document for the full list of supported options
+  async ({ body, client }) => { // The second argument is the handler function itself
+    console.log('Incoming suggestion handler invocation', body);
+    // Fetch some inspirational quotes
+    const apiResp = await fetch("https://motivational-quote-api.herokuapp.com/quotes");
+    const quotes = await apiResp.json();
+    console.log('Returning', quotes.length, 'quotes');
+    const opts = {
```
seratch, line 154: I forgot the exact number but we may want to mention the maximum number of options/option_groups here for more developer-friendliness

src/functions/types.ts
```diff
@@ -349,6 +350,12 @@ export type SlackFunctionType<Definition> = Definition extends
           FunctionDefinitionArgs<I, O, RI, RO>
         >,
       ): SlackFunctionType<Definition>;
+      addBlockSuggestionHandler(
```
seratch, line 353: I know, if we add this change, we should consistently update other handler definitions too but what do you think about something like this? (disclaimer: the diff is incomplete - just for demo purpose)

```diff
diff --git a/src/functions/interactivity/types.ts b/src/functions/interactivity/types.ts
index 0c9bd30..6d17983 100644
--- a/src/functions/interactivity/types.ts
+++ b/src/functions/interactivity/types.ts
@@ -28,8 +28,9 @@ export type BlockSuggestionHandler<Definition> = Definition extends
   FunctionDefinitionArgs<infer I, infer O, infer RI, infer RO> ? {
     (
       context: SuggestionContext<FunctionRuntimeParameters<I, RI>>,
-      // deno-lint-ignore no-explicit-any
-    ): Promise<any> | any;
+    ): Promise<{ options: [] } | { option_groups: [] }> | { options: [] } | {
+      option_groups: [];
+    };
   }
   : never;
 
diff --git a/src/functions/slack-function_test.ts b/src/functions/slack-function_test.ts
index fd81222..4cf0051 100644
--- a/src/functions/slack-function_test.ts
+++ b/src/functions/slack-function_test.ts
@@ -110,7 +110,11 @@ Deno.test("addBlockSuggestionHandler", async () => {
 
   const actionId = "whatever";
   const suggestionSpy = mock.spy();
-  typedHandlers.addBlockSuggestionHandler(actionId, suggestionSpy);
+  typedHandlers.addBlockSuggestionHandler(actionId, async ({ inputs }) => {
+    console.log(inputs);
+    // return await { options: [] };
+    return await { option_groups: [] };
+  });
 
   await typedHandlers.blockSuggestion({ body: { action_id: actionId } });
   mock.assertSpyCalls(suggestionSpy, 1);
```

docs/functions-suggestion-handlers.md
```diff
@@ -0,0 +1,218 @@
+## Block Kit Suggestion Handlers
+
+Your application's [functions][functions] can do a wide variety of interesting
+things: post messages, create channels, or anything available to developers via
+the [Slack API][api]. One of the more compelling features available to app developers
+is the ability to use [Block Kit][block-kit] to add richness and depth to messages
+in Slack. Even better, [Block Kit][block-kit] supports a variety of [interactive components][interactivity]!
```
vegeris, line 7: :twocents: I was wondering about built-ins while reading this (normally wouldn't call the Slack API unless I couldn't use a built-in)

`Your application can use our built-in functions to interact with the Slack API, or you can call our endpoints directly within your custom functions. One reason you might to do that is to add extra formatting and interactivity to messages sent through your app. [Does Schema.slack.functions.SendMessage support displaying messages with Block Kit as well?]`

filmaj: I think the sendmessage built-in does indeed support block kit payloads: https://github.com/slackapi/deno-slack-sdk/blob/main/src/schema/slack/functions/send_message.ts#L42

docs/functions-suggestion-handlers.md
filmaj: I couldn't find a reference to what is the maximum number of options one can return... I did end up posting in our api docs channel about this issue: https://slack-pde.slack.com/archives/C8DPMVAQY/p1666275726696799

src/functions/types.ts
filmaj: @seratch I made another pass at adding block-suggestion handler-specific return types. I think it is better now. Can you tell me what you think?

docs/functions-suggestion-handlers.md
vegeris: (Unrelated to this PR but) Found [this](https://slack-pde.slack.com/archives/C01FA1FMGE9/p1663719958337829?thread_ts=1663649516.729709&cid=C01FA1FMGE9) helpful thread; Rich Text block or text accepted in the SendMessage message property and button type blocks accepted in the interactive_blocks property. No wonder the section block I was trying to pass kept erroring out

src/functions/interactivity/suggestion_router.ts
```diff
@@ -0,0 +1,189 @@
...
+  matchBasicConstraintField,
+  normalizeConstraintToArray,
+} from "./matchers.ts";
+
+/**
+ * Define a suggestion "router" and its input and output parameters for use in a Slack application. The SuggestionRouter will route incoming action events to action-specific handlers.
+ * @param {SlackFunctionDefinition<InputParameters, OutputParameters, RequiredInput, RequiredOutput>} func Reference to your previously-created SlackFunction, defined via DefineFunction
+ * @returns {SuggestionRouter}
+ */
+export const BlockSuggestionRouter = <
```
shapirone, line 26: This isn't exported from `deno-slack-sdk/mod.ts`, but `BlockActionsRouter` is. I bring this up for
1. Consistency
2. The jsdoc comments here are pretty extensive, but I'm not sure they're being surfaced anywhere (I'm using `SlackFunction().addBlockSuggestionHandler()` but we don't surface this directly.

What do you think? Should we export from the top level or alternatively should we deprecate `BlockActionsRouter` being exported from there?

src/functions/interactivity/types.ts
```diff
@@ -22,6 +24,60 @@ export type BlockActionHandler<Definition> = Definition extends
...
+   * @description An array of {@link MenuOption} objects that belong to this specific group. Maximum of 100 items.
+   */
+  options: MenuOption[];
+};
+
+type MenuOption = {
+  /**
+   * @description A {@link PlainTextObject} that defines the text shown in the option on the menu. Maximum length for the `text` property inside this field is 75 characters.
+   */
+  text: PlainTextObject;
```
shapirone, line 62: This only supports PlainText and doesn't support something like mrkdwn? I think that makes sense, just sanity checking

shapirone (approved):
Extremely well documented in md and in jsdoc with great test coverage. Thank you Fil! Left a couple questions/musings and nits, but nothing blocking.

src/functions/types.ts
seratch: @filmaj I don't remember both plain_text and mrkdwn formats are supported for text, label but apart from that, it looks great!

src/functions/interactivity/block_suggestion_types.ts
```diff
@@ -0,0 +1,160 @@
+import type { BlockAction, BlockElement } from "./block_kit_types.ts";
```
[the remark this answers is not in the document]
filmaj: Not really! Do you have a preference?

src/functions/interactivity/suggestion_router.ts
filmaj: Good callout. So I have a [PR up](https://slack-github.com/slack/webapp/pull/158071) to update the interactivity docs on api.slack.com to not use these Router constructs but rather the `add*Handler` methods hanging off the top-level function. Once that is merged, I think we can then go through the motions to deprecate / no longer export the Router bits and, probably at that time it would be appropriate, also refactor this entire mess to DRY up the interactivity routing for all interactivity events.

src/functions/interactivity/types.ts
filmaj: That's correct. See the `text` Field reference in this doc: https://api.slack.com/reference/block-kit/composition-objects#option

> A [text object](https://api.slack.com/reference/block-kit/composition-objects#text) that defines the text shown in the option on the menu. **Overflow, select, and multi-select menus can only use plain_text objects**, while radio buttons and checkboxes can use mrkdwn text objects. Maximum length for the text in this field is 75 characters.

src/functions/types.ts
filmaj: According to [the reference docs (under Fields)](https://api.slack.com/reference/block-kit/composition-objects#option), only plain_text is supported for select / dropdown / overflow menus:

> A [text object](https://api.slack.com/reference/block-kit/composition-objects#text) that defines the text shown in the option on the menu. Overflow, select, and multi-select menus can only use plain_text objects, while radio buttons and checkboxes can use mrkdwn text objects. Maximum length for the text in this field is 75 characters.

docs/functions-suggestion-handlers.md
```diff
@@ -0,0 +1,218 @@
+## Block Kit Suggestion Handlers
```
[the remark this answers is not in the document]
filmaj: Done!

src/functions/interactivity/block_suggestion_types.ts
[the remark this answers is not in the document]
filmaj: @shapirone should I then rename the action_router to block_action_router, too?

shapirone (approved):

PR #122, opened 2022-12-07
Adding more tests to ParameterVariable

filmaj:
... to exercise use of a Custom Type, based on a Typed Object, with different values for additionalProperties.


shapirone (approved):
I think this is fundamentally testing the same thing that our object tests cover w/r/t `additionalProperties`, but I'm okay with the little bit of redundancy here especially if we want additional confidence in the big changes we might need to be making

PR #123, opened 2022-12-09
Default malformed workflow step inputs to `undefined` so that the manifest validation method can return useful errors

jesselawson:
###  Summary

Errors that occur while reading a malformed manifest are rather ambiguous. 

Here I am trying to run `slack run` and getting some errors that are difficult for me to interpret:

manifest-before

If we wrap the call to `JSON.parse` [here](https://github.com/slackapi/deno-slack-sdk/blob/main/src/workflows/workflow-step.ts#L48) in a `try-catch` and set the templatized input to `undefined` in the event of an exception, the CLI will return the (helpful) manifest validation results from `/api/apps.manifest.validate`.

After testing this change in my branch [here](https://github.com/jesselawson/deno-slack-sdk/blob/jesse_better_manifest_error_reporting/src/workflows/workflow-step.ts#L48-L51), then updating my app to use my local fork of the sdk, the above command now yields this:

manifest-after

I also added a deno-lint-ignore comment for the empty catch block. 

My hopes are this can start a discussion about better error reporting to improve the developer troubleshooting experience. 

```
$ slack version
Using slack v1.17.0

$ slack doctor
...
✔ Dependencies 
        deno_slack_hooks: 0.8.0
        deno_slack_sdk: 1.4.3
        deno_slack_api: 1.5.0
```

* [X] I've read and understood the [Contributing Guidelines](https://github.com/{project_slug}/blob/main/.github/CONTRIBUTING.md) and have done my best effort to follow them.


deno.lock
```diff
@@ -0,0 +1,60 @@
+{
```
filmaj, line 1: Ooooo lockfile! This is a new one for me in a deno project. Should we have one of these now?

src/workflows/workflow-step.ts
```diff
@@ -45,7 +45,11 @@ abstract class BaseWorkflowStepDefinition {
       {} as ManifestWorkflowStepSchema["inputs"];
 
     for (const [inputName, inputValue] of Object.entries(this.inputs)) {
-      templatizedInputs[inputName] = JSON.parse(JSON.stringify(inputValue));
+      try {
+        templatizedInputs[inputName] = JSON.parse(JSON.stringify(inputValue));
+      } catch {
+        templatizedInputs[inputName] = undefined;
```
filmaj, line 51: So if I understand correctly, setting this to `undefined` will initiate a failure mode because the backend manifest validation API will reject the manifest?

filmaj (commented):
Thanks for tackling this! Left two questions: one around the deno lockfile, and one around how this PR works / the failure mode we want to trigger.

deno.lock
shapirone: I haven't spent too much time thinking about this. Right now I'm leaning towards us including it, but @jesselawson if you could remove this from this PR I think that will make approving this one easier!

jesselawson: Yes I'll remove it -- wanted to keep it here to make sure it's on _someone's_ radar.

src/workflows/workflow-step.ts
jesselawson: That's how I understand it, but I haven't dug into the internals of the validation API. 

What I am observing is:

In cases where JSON.stringify returns `undefined`, JSON.parse [throws an exception](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions). 

When JSON.parse throws an exception, the Deno interpreter dies (a design decision of Deno was to [always die on uncaught exceptions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions)), bubbling the TypeScript interpreter's exit issue (the uncaught `SyntaxError` from JSON.parse) to the developer, instead of allowing the manifest validation API to do its job and return manifest validation issues to the developer.

deno.lock
jesselawson: [Removed](https://github.com/slackapi/deno-slack-sdk/pull/123/commits/09b68c60869fbf3255956ca7f012bb150a2b64b6)

src/workflows/workflows_test.ts
```diff
@@ -173,3 +173,37 @@ Deno.test("Workflows properly treats interactivity and user context types", () =
     `{{steps.0.user.secret}}`,
   );
 });
+
```
jesselawson, line 176: I don't know if it's common to document _why_ certain tests exist, but in this case, I can see the value in a comment that explains what I wrote [here](https://github.com/slackapi/deno-slack-sdk/pull/123#discussion_r1046372790).

Something like:

Deno will exit on uncaught exceptions. JSON.parse will raise an exception when given `undefined`. We can set `undefined` manually here in the case of an `undefined` inputValue, which will allow Deno to continue and pass the undefined values up to the validation API -- which will communicate back to the CLI the specific validation errors it ran in to. 

Or something like that.

shapirone (approved):
This seems good to me. There is a small concern we may be ignoring and therefore unaware that one of the other [JSON.parse `SyntaxError` cases](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/JSON_bad_parse) is occurring here, but those seem much less likely immediately after running `JSON.stringify` on a JS object. Even if this occurs, they'll still hit an error at manifest validation and will point them to the right area, but it may not be clear to the developer may why that value is missing. If we did want to continue to throw on non-undefined cases, we could add logic to see if `JSON.stringify` results are `undefined` before running `JSON.parse`.

Eventually it might be nice for us to add some kind of logging (verbose maybe) for discoverability or so that we can track these errors, but we can't have that until we allow support for writing to stdout at manifest generation time.

Thanks again Jesse!

src/workflows/workflows_test.ts
shapirone: That sounds good to me!

I really like this approach to documenting/explaining test cases as it keeps the knowledge within the codebase but doesn't introduce any bloat or distraction into the codepath. I think we should do this elsewhere as well, maybe as we do our exercise @filmaj ?

filmaj: 100% useful, I try to do that as much as possible myself (esp. via test titles, e.g. https://github.com/slackapi/deno-slack-sdk/compare/main...param-tests-custom-types-moarprops#diff-9dc41c90f103a2563c551a62f729231125264b51509240a83e38117f4e50feb7R62)

src/workflows/workflows_test.ts
```diff
@@ -173,3 +173,46 @@ Deno.test("Workflows properly treats interactivity and user context types", () =
...
+/*
+  Deno will exit on uncaught exceptions. 
+  JSON.parse will raise an exception when given undefined. 
+  By setting undefined manually to reflect the case of an 
+  undefined inputValue, which will allow Deno to continue 
+  and pass the undefined values up to the validation API 
+  -- which will then communicate back to the CLI the specific 
+  validation errors it ran in to.
+*/
```
jesselawson, line 185: @shapirone 👀 🤔

jesselawson: Sorry, it looks like the Deno tests are failing because of the multi-line comments: 

Screen Shot 2022-12-13 at 9 52 11 AM

I'll switch it to `//`.

jesselawson: That seems to [have worked](https://github.com/slackapi/deno-slack-sdk/actions/runs/3688011788/jobs/6242303353).

PR #124, opened 2022-12-15
Support for optional/required properties on nested object parameters

filmaj:
###  Summary

This is related to HERMES-4071.

I feel like working on this has aged me by like 3 years.

Summary of user-facing changes:

- Recommend wrapping any typed-object parameters in a new `DefineProperty` helper function

Summary of more internal changes:

- Creating `ParameterDefinitionWithGenerics` as a generic version of `ParameterDefinition`
- Passing generics to `TypedObjectParameterDefinition` so that a constraint exists between `properties` and `required`
  - Creating `TypedObjectParameter` as a way to pass around the generic-less schema/shape for reference in `extends` constraints
- Move from using TS `type` to `interface` so that JSDoc comments are inherited appropriately

# TODO

- [x] `DefineParameter` name OK, or something else?


src/dev_deps.ts
```diff
@@ -8,5 +8,24 @@ export {
...
+ */
+export type CannotBe<T, U> = Extract<T, U> extends never ? true : false;
+/**
+ * Checks whether the provided type parameter allows to be undefined.
+ * Useful for checking optionality.
+ */
+export type CanBeUndefined<T> = CanBe<T, undefined> extends true ? true
+  : false;
+export type CannotBeUndefined<T> = CannotBe<T, undefined> extends true ? true
+  : false;
```
shapirone, line 31: I'm thinking we might want to include these in our `src/test_utils.ts` folder instead

filmaj: It's been a while - what's your opinion on this now @shapirone ?

src/functions/types.ts
```diff
@@ -70,29 +74,33 @@ type FunctionInputRuntimeType<
...
-      >
-    : Param["type"] extends typeof SchemaTypes.string ? string // Not a Custom Type, so assign the runtime value
+    : Param["type"] extends ICustomType
+      ? Param extends CustomTypeParameterDefinition ? FunctionInputRuntimeType<
+          Param["type"]["definition"],
+          IncreaseDepth<CurrentDepth>
+        >
+      : UnknownRuntimeType
+    : Param["type"] extends
+      | typeof SchemaTypes.string
```
filmaj, line 84: This particular change is just grouping all the types that yield a `string`

src/dev_deps.ts
shapirone: I feel the same, `dev_deps` is intended for importing so AFAIK we shouldn't be defining our own things in here 😅

src/functions/types.ts
```diff
@@ -70,29 +74,33 @@ type FunctionInputRuntimeType<
...
+      | typeof SlackSchemaTypes.user_id
+      | typeof SlackSchemaTypes.usergroup_id
+      | typeof SlackSchemaTypes.channel_id
+      | typeof SlackSchemaTypes.date ? string // Not a Custom Type, so assign the runtime value
     : Param["type"] extends
       | typeof SchemaTypes.integer
-      | typeof SchemaTypes.number ? number
+      | typeof SchemaTypes.number
```
filmaj, line 91: Same here for `number`

src/functions/types.ts
```diff
@@ -101,21 +109,24 @@ type FunctionInputRuntimeType<
 type UnknownRuntimeType = any;
 
 type TypedObjectFunctionInputRuntimeType<
-  Param extends TypedObjectParameterDefinition,
-> = Param["additionalProperties"] extends false ? {
-    [k in keyof Param["properties"]]: FunctionInputRuntimeType<
-      Param["properties"][k]
+  Props extends TypedObjectProperties,
```
filmaj, line 112: This is the main change enabling proper object sub-property optionality autocomplete in a function implementation context.

src/parameters/define_parameter.ts
```diff
@@ -0,0 +1,15 @@
+import {
+  TypedObjectParameterDefinition,
+  TypedObjectProperties,
+  TypedObjectRequiredProperties,
+} from "../parameters/types.ts";
+
+export const DefineParameter = <
```
filmaj, line 7: The only purpose this method serves is to create a 'hook' for the type system to be able to apply the two generics below this line, and in particular, that is important because the generics sequentially rely on each other; that is, the third generic relies on the previous two named generics, and the second generic relies on the first named generic.

src/parameters/define_parameter_test.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+    },
+    required: ["aString"],
+    additionalProperties: true,
+  });
+  assert<IsExact<typeof obj.required, "aString"[]>>(true);
+});
+
+/*
+TODO: DefineParameter fails to constrain the required field to property names :(
```
filmaj, line 23: Really wish we could solve this issue 😢

src/parameters/mod.ts
```diff
@@ -21,46 +37,54 @@ export type PossibleParameterKeys<
...
   required: Required;
 };
 
 export type ParameterVariableType<
   Def extends ParameterDefinition,
-> = Def["type"] extends ICustomType // If the ParameterVariable is a Custom type, use it's definition instead
-  ? ParameterVariableType<Def["type"]["definition"]>
-  : Def extends TypedObjectParameterDefinition // If the ParameterVariable is of type object, allow access to the object's properties
-    ? ObjectParameterVariableType<Def>
+  CurrentDepth extends RecursionDepthLevel = 0,
```
filmaj, line 49: It's a complete mystery to me why previously, with no recursion guard whatsoever, the old code didn't lead to an infinite-recursion TS error.

src/schema/slack/functions/open_form.ts
```diff
@@ -27,12 +27,9 @@ export default DefineFunction(
         },
         fields: {
           type: InternalSlackTypes.form_input_object,
-          description: "Input fields to be shown on the form",
```
filmaj, line 30: Huh, should these be removed?

src/schema/slack/types/custom/form_input.ts
```diff
@@ -1,8 +1,9 @@
 import SchemaTypes from "../../../schema_types.ts";
 import { DefineType } from "../../../../types/mod.ts";
+import { DefineParameter } from "../../../../parameters/define_parameter.ts";
 
-const FormInput = DefineType({
-  name: "slack#/types/form_input_object",
+const FormInputObject = DefineParameter({
```
filmaj, line 5: I think _technically_ we can avoid using `DefineParameter` here as I believe the call to `DefineType` below should function in the same way in terms of providing an entrypoint for the props/required props generics to come into play.

That said, this is all internal stuff not directly exposed to the dev, and I'd rather we use `DefineParameter` to wrap all typed object definitions whenever we can, even if redundant.

Open to either. This same comment applies to all three of the Slack custom types.

src/functions/tester/mod.ts
```diff
@@ -42,6 +42,7 @@ export const SlackFunctionTester: SlackFunctionTesterFn = <
     const ts = new Date();
     const token = args.token || "slack-function-test-token";
 
+    // TODO: can we reuse some of our existing types for modeling payloads to ensure this structure doesnt become out of date?
```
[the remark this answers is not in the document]
shapirone: The `SlackFunctionTester` needs to be in `src` so that it's shared with developers, no? It's what allows them to run unit tests on their functions.

src/functions/types.ts
```diff
@@ -47,13 +50,14 @@ export type FunctionInvocationBody = {
 };
 
 /** @description Defines accepted depth values */
-type RecursionDepthLevel = 0 | 1 | 2 | 3 | 4 | 5;
+export type RecursionDepthLevel = 0 | 1 | 2 | 3 | 4 | 5;
 
 /** @description Defines the max depth we want to recurse */
-type MaxRecursionDepth = 5;
+export type MaxRecursionDepth = 5;
```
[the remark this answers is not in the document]
shapirone: Ah yeah, might be good to move recursion guard logic to a shared space then!

src/functions/tester/mod.ts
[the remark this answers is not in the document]
filmaj: Damn I keep forgetting that this is exposed to devs! Good call, resolving.

src/functions/types.ts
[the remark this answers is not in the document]
filmaj: Hmm, maybe `src/types.ts`? `src/shared_types.ts`?

src/parameters/define_parameter_test.ts
shapirone: Agreed, the solution we found leads to another inexplicable issue where devs may need to instantiate a variable that uses `DefineParameter` outside of a `DefineFunction` call, which seems unacceptable to me

src/schema/slack/types/custom/form_input.ts
shapirone: Good catch, I'd like for us to just use `DefineType` rather than spreading the result of `DefineParameter` into a `DefineType` call, since I think that's what we would recommend to devs.

filmaj: K I updated this in https://github.com/slackapi/deno-slack-sdk/pull/124/commits/26345ff5ce097f1c06efd90fa96dd9713933bd2d and also added a test for the internal `FormInput` slack type to make sure it behaves as we would expect in a function runtime context.

src/schema/slack/functions/open_form.ts
WilliamBergamin: This will eventually be added back when the script is ran again, but these are generated changes

shapirone: Do we know why or how these made it into the PR? Was it me? 😅

WilliamBergamin: Changing the contents of the `function.json` could easily have caused a change in the generated files

shapirone: But someone would have had to run the generation script to change these files. IMO we shouldn't commit these changes

filmaj: It was hard for me to figure out where this blame came from haha. Should I just put these descriptions back ?

src/functions/types.ts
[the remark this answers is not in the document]
shapirone: Something like `src/type_utils.ts` would match `src/test_utils.ts` 🤷

src/manifest/manifest_schema.ts
```diff
@@ -1,9 +1,7 @@
 import { ISlackDatastore } from "../datastore/types.ts";
 import { ISlackFunctionDefinition } from "../functions/types.ts";
-import type {
-  ParameterDefinition,
-  ParameterSetDefinition,
-} from "../parameters/mod.ts";
+import type { ParameterSetDefinition } from "../parameters/mod.ts";
```
shapirone, line 3: ultra nit: The fact that we didn't move this to `parameters/types.ts` as well sticks out to me here

src/manifest/manifest_test.ts
```diff
@@ -32,6 +32,7 @@ Deno.test("SlackManifestType correctly resolves to a Hosted App when runOnSlack
     icon: "icon.png",
     botScopes: ["channels:history", "chat:write", "commands"],
   };
+
```
shapirone, line 35: All changes in this file seem unnecessary, my best guess is that they're remnants of when we didn't support `required` being `undefined` (w/ some formatting changes)

src/functions/types.ts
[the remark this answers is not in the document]
filmaj: Aight done in fb96d03

src/parameters/define_parameter_test.ts
```diff
@@ -0,0 +1,40 @@
+import { DefineProperty } from "./define_parameter.ts";
```
shapirone, line 1: Need to rename the file we're importing from, and this test file

src/parameters/mod.ts
```diff
@@ -21,46 +37,54 @@ export type PossibleParameterKeys<
...
+type ObjectParameterVariableType<
+  Def extends TypedObjectParameter,
+> = Def["additionalProperties"] extends false
+  ? ObjectParameterPropertyTypes<Def["properties"]>
+  : 
+    & ObjectParameterPropertyTypes<Def["properties"]>
+    & {
+      // deno-lint-ignore no-explicit-any
+      [key: string]: any;
+    };
```
shapirone, line 87: Nit: We could do what we did for runtime functions and remove the need to specify `ObjectParameterPropertyTypes<Def["properties"]>` twice

```suggestion
type ObjectParameterVariableType<
  Def extends TypedObjectParameter,
> =
  & ObjectParameterPropertyTypes<Def["properties"]>
  & (Def["additionalProperties"] extends false ? Record<never, never>
    : {
      // deno-lint-ignore no-explicit-any
      [key: string]: any;
    });
```

src/parameters/mod.ts
```diff
@@ -69,34 +93,42 @@ export const ParameterVariable = <P extends ParameterDefinition>(
...
-      param = CreateUntypedObjectParameterVariable(namespace, paramName);
-    }
-  } else {
-    param = CreateSingleParameterVariable(
-      namespace,
-      paramName,
-    ) as ParameterVariableType<P>;
   }
-
+  // TODO: the following type assertion seems to ignore the case where param could be null
```
shapirone, line 121: Can we do anything for this todo?

src/parameters/mod.ts
```diff
@@ -69,34 +93,42 @@ export const ParameterVariable = <P extends ParameterDefinition>(
...
+    return ParameterVariable(
       namespace,
       paramName,
       definition.type.definition,
     );
-  } else if (definition.type === SchemaTypes.object) {
-    if ("properties" in definition) {
-      param = CreateTypedObjectParameterVariable(
+  }
+  switch (definition.type) {
```
shapirone, line 103: 💭 Early on in our exploration we moved to this `switch` statement. Do we think that using a switch statement is still preferable considering we needed to add an if statement above?

src/manifest/manifest_schema.ts
filmaj: Done in 54e5989

src/manifest/manifest_test.ts
filmaj: Fixed in d09eb32

src/parameters/define_parameter_test.ts
filmaj: Fixed in 9b73d87

src/parameters/mod.ts
filmaj: Can we chalk it up as something to tackle later and not block this PR anymore? I can't fully remember what the coment is about to be honest

src/schema/slack/functions/open_form.ts
shapirone: Yes please, none of our built-ins should have changes in this PR

src/parameters/mod.ts
filmaj: Probably doesn't make any sense anymore. Should I remove it and flip back to `if`s?

shapirone: Yeah that makes sense to me

src/schema/slack/types/custom/form_input.ts
```diff
@@ -18,6 +19,7 @@ const FormInput = DefineType({
       },
     },
   },
+  required: [],
```
shapirone, line 22: I think both of these are required, but we may need to double check

src/schema/slack/functions/open_form.ts
filmaj: Done in bdbdce8

src/schema/slack/types/custom/form_input.ts
filmaj: How/where can I double check?

src/parameters/mod.ts
shapirone: Heh, I also didn't follow what the comment was concerned about which is why I was hoping we could resolve before merging 😅 Fixing it by clarifying or removing isn't a PR blocker but would be nice to resolve rather than leaving it in here forever without clarifying

filmaj: I removed it in 30d8fef - the initializing variable at the top of this method initially assigns `null` to it but with the way the logic works out now, it will never remain `null`, so I think it is safe to remove.

src/schema/slack/types/custom/form_input.ts
shapirone: I checked by testing an app that uses `OpenForm` (which uses `FormInput`) and commented out both `elements` and `required`. Turns out `elements` is required, but `required` is optional! Updated accordingly

shapirone (approved):
lgtm

PR #127, opened 2023-01-26
Function generation script improvements

WilliamBergamin:
###  Summary

This is related to HERMES-3364

The goal of this PR is to:
1. Add unit tests to the existing generated built in TypeScript functions
2. improve the generation script to remove generated TypeScript functions that no longer exist (don't delete the unit tests, let the devs do that)
3. Improve the script to generate unit tests for generated TypeScript functions that do not have tests.

NOTE the Typescript functions are not up to date with the API I did not include this in the PR


src/schema/slack/functions/_scripts/README.md
```diff
@@ -1,10 +1,12 @@
...
-`schema/slack/functions` directory, i.e.
-`schema/slack/functions/send_message.ts`. It will also update the
+This script will generate the necessary function TypeScript files along with
+their tests in the `schema/slack/functions` directory, i.e.
+`schema/slack/functions/send_message.ts` and
+`schema/slack/functions/send_message_test.ts`. It will also update the
 `schema/slack/functions/mod.ts` file to import/export all of the defined
-functions.
+functions. It will also remove outdated function TypeScript files but not their
+corresponding test, the tests must be removed manually.
```
shapirone, line 9: What's the purpose of making them be removed manually? The long term desire is to have all of this automated to the point where it can auto-open a PR in our repo when a built-in function is changed on the backend

WilliamBergamin: Thanks for bringing this up, I did move fast on this PR and maybe should have consulted you before, I did not know the goal of this was to make it auto-open PRs (great idea 🥇 )

Removing one of these TypeScript built-in `functions` can introduce a breaking change for our users. My reasoning here was to prevent the script from  accidentally removing a `function` along with its `test` making it harder to know what was removed, if the `test` stays and then fails we will know that something is gone and can confirm it is intentional

At the same time in a PR we should be able to see clearly what was deleted and if this is intentional

shapirone: Ah, definitely see where you're coming from!

> At the same time in a PR we should be able to see clearly what was deleted and if this is intentional

While I think this is the right moment for us to question/block a breaking change like this, it's also too late for us to do anything about because the backend's support will have already been removed and any use of the built-in function will already be broken e2e. We'll at least know to make this merge a breaking SDK change and therefore a major version bump.

shapirone: Ah, definitely see where you're coming from!

> At the same time in a PR we should be able to see clearly what was deleted and if this is intentional

While I think this is the right moment for us to question/block a breaking change like this, it's also too late for us to do anything about because the backend's support will have already been removed and any use of the built-in function will already be broken e2e. We'll at least know to make this merge a breaking SDK change and therefore a major version bump.

WilliamBergamin: I see 👁️ I can add test deletion to the script in that case, I originally had it in.

I have a follow up question, should we also add more complex `test` generation? currently the generated `tests` will fail by default, someone needs to go and manually make them pass. I would tend not to update existing `tests` that way breaking changes can be detected

PR #130, opened 2023-02-01
Add format property to StringParameterDefinition type

srajiang:
###  Summary

Adds `format` property to StringParameterDefinition type.

## Testing
You can easily clone down this app and follow these [instructions](https://github.com/srajiang/test-format-string/blob/7926cb0d803dc930350d72812b1dde49cb1a485e/functions/greeting_function.ts#L24-L32) to test out the addition. 

Errors you should see: 
1. Manifest validation error when value for `format` does not match either "email" or "url". For example `uri` should both cause a manifest validation error. 
Screenshot 2023-02-01 at 5 14 13 PM

2. Parameter validation error when the value does not conform to the specified format
Screenshot 2023-02-01 at 5 14 46 PM

### DIFF
--- a/src/parameters/types.ts
+++ b/src/parameters/types.ts
@@ -79,6 +79,8 @@ type StringParameterDefinition = BaseParameterDefinition<string> & {
   enum?: string[];
   /** Defines labels that correspond to the `enum` values. */
   choices?: EnumChoice<string>[];
+  /** Define accepted format of the string */
+  format?: "url" | "email";
 };
 
 type IntegerParameterDefinition = BaseParameterDefinition<number> & {


filmaj (approved):
Looks fine to me.

One thing I find weird is that there are no static errors for providing an incorrect value to the `format` field. I even tried rebasing this branch on our latest `DefineParameter` work, but that didn't help either (though I think that has to do with the fact `DefineParameter` currently only supports `Schema.types.object` and not `string`).

@shapirone WDYT? I think we probably want to note this in our post-`DefineParameter` refactor work to see if we can raise static errors for providing invalid values to `format` (and maybe this issue also exists for other string type properties like `choices`, `enum`, etc).

PR #134, opened 2023-02-03
Split up function runtime context integration tests into multiple files

filmaj:
###  Summary

Totally optional, feel free to close if you don't like it @shapirone .

Also tweaked `deno.jsonc` to lint/fmt check the test files too. The last commit in this PR encapsulates those more config/linting changes, FYI.

Thought it felt nice, lemme know WYT.


tests/integration/functions/runtime_context/string_parameters_test.ts
```diff
@@ -0,0 +1,156 @@
...
+      outputs: {},
+    };
+  };
+  const { createContext } = SlackFunctionTester(TestFn);
+  const inputs = { in: "test" };
+  const result = handler(createContext({ inputs }));
+  assertEqualsTypedValues(result.outputs, {});
+});
+
+Deno.test("Custom Function with a string input and output should provide the string input in a function handler context", () => {
```
shapirone, line 43: Just noticing the name for this test mentions the string input in context, but doesn't really touch on the output which is what distinguishes this test from the one above. Should this test even care about the `input` if that's covered by the above test?

tests/integration/functions/runtime_context/string_parameters_test.ts
```diff
@@ -0,0 +1,156 @@
...
+import { CanBe, CanBeUndefined } from "../../../../src/test_utils.ts";
+import { DefineFunction } from "../../../../src/mod.ts";
+import {
+  EnrichedSlackFunctionHandler,
+} from "../../../../src/functions/types.ts";
+import { SlackFunctionTester } from "../../../../src/functions/tester/mod.ts";
+import { assertEqualsTypedValues } from "../../../../src/test_utils.ts";
+
+/**
+ * Custom Function handler tests, exercising string-based inputs and outputs
```
shapirone, line 13: Something that's interesting to me about these tests is that I feel like the intent of the tests is based on the functionality of the handler (defined inputs are available, defined outputs are expected), and less focused on the type of parameter. I think this is supported by the fact that we don't have tests for the other simple types (boolean, number, channel_id, etc.), but maybe we should.

Should we have this file focus on handler functionality like the inputs/outputs/optionality? Should we add tests for our other primitive types?

tests/integration/functions/runtime_context/typed_object_parameters_test.ts
```diff
@@ -0,0 +1,420 @@
+import {
```
shapirone, line 1: File named typed object parameter or typed object property?

tests/integration/functions/runtime_context/untyped_object_parameters_test.ts
```diff
@@ -0,0 +1,64 @@
+import {
```
shapirone, line 1: same ❔ from typed objects

tests/integration/functions/runtime_context/custom_type_parameters_test.ts
```diff
@@ -1,38 +1,32 @@
...
+ * Custom Function handler tests, exercising Custom Type inputs/outputs, including built-in/internal Slack custom types
  */
-
-
-// TODO: unwrapped typed object, in a custom type, fed to a typed array, yields an array of `any` items at runtime
-/*
-Deno.test("Custom Function using Typed Arrays of Custom Types of unwrapped typed objects should honor required and optional properties", () => {
-  const obj = {
+Deno.test("Custom Function using a Custom Type input for a DefineProperty-wrapped typedobject with mixed required/optional properties should provide correct typedobject typing in a function handler context and should complain if typedobject output does not include required property", () => {
+  const myObject = DefineProperty({
```
shapirone, line 18: Should we be using `DefineProperty` for Custom Type examples?

shapirone (changes requested):
Mostly dropping some comments, I think the biggest one to resolve before merging is the name/purpose of the `failure_incomplete_test.ts` file

tests/integration/functions/runtime_context/failure_incomplete_test.ts
```diff
@@ -0,0 +1,114 @@
...
+  const handler: EnrichedSlackFunctionHandler<typeof TestFn.definition> =
+    () => {
+      return {
+        outputs: {},
+      };
+    };
+  const { createContext } = SlackFunctionTester(TestFn);
+  const result = handler(createContext({ inputs: {} }));
+  assertEqualsTypedValues(result.outputs, {});
+});
```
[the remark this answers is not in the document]
filmaj: Could split out the empty/undefined parameter tests into a separate file and keep this file for failure mode tests. Also unclear to me where exercising the `complete` flag would fall into.. any ideas?

tests/integration/functions/runtime_context/string_parameters_test.ts
filmaj: Good call. Taking a closer look, maybe we should have dedicated tests for the inputs which assert that within a runtime context, the input is typed properly. For the outputs, we would assert that not providing the correct output name and type raises an error.
WDYT?

tests/integration/functions/runtime_context/failure_incomplete_test.ts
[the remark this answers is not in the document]
shapirone: That makes sense to split the empty/undefined tests into their own file. I think having incomplete or error in the same file makes sense as it's the alternative to a successfully completed output which is what most of our tests cover. Maybe rather then calling it fail, we refer to it as error? So something like `incomplete_error_status_test.ts`?

tests/integration/functions/runtime_context/string_parameters_test.ts
filmaj: I like focusing the tests on handler functionality like input, and output and optionality. As for how to organize them, some ideas, wondering what your thoughts are:

- could name the test files based on the type, and exercise inputs, outputs and optionality per type. So we'd keep the file name for this one as-is, but I could add a `boolean_parameters_test.ts`, `number_parameters_test.ts`, etc.
- could name the test files based on functionality (`input_parameters_test.ts`, `output_parameters_test`, `input_parameter_optionality_test.ts`) and exercise all primitive (and complex?) types in them
- could name the test files based on both! `string_input_parameters_test.ts`, `string_output_parameters.ts`, `string_input_parameter_optionality_test.ts`, etc. etc. Lotsa files, but presumably they would be tiny (maybe even only one or two tests per file).

Just some ideas off the top of my head. Lemme know what you think!

tests/integration/functions/runtime_context/typed_object_parameters_test.ts
filmaj: Ooof, tough one haha. Either works for me. What do you prefer?

tests/integration/functions/runtime_context/custom_type_parameters_test.ts
filmaj: No, we shouldn't, but I wanted to make sure it works!

Should I remove them?

tests/integration/functions/runtime_context/typed_object_parameters_test.ts
shapirone: Property since it's consistent with the method name

tests/integration/functions/runtime_context/custom_type_parameters_test.ts
shapirone: Yes please!

tests/integration/functions/runtime_context/failure_incomplete_test.ts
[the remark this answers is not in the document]
filmaj: Done in 6d8fd23

tests/integration/functions/runtime_context/string_parameters_test.ts
shapirone: I think let's start with option 2 — split into `input`, `output`, and `optionality`.

Then, either in this PR or another PR we could introduce type specific tests. I don't think we need to have exhaustive testing on types, just ensuring that at runtime a string is a string, a boolean is a boolean, etc.

tests/integration/functions/runtime_context/typed_object_parameters_test.ts
filmaj: Done in 9e1f9fd

tests/integration/functions/runtime_context/untyped_object_parameters_test.ts
filmaj: Done in 9e1f9fd

tests/integration/functions/runtime_context/string_parameters_test.ts
shapirone: Yeah, this feels inline with our [other thread here](https://github.com/slackapi/deno-slack-sdk/pull/134#discussion_r1097784986)

tests/integration/functions/runtime_context/custom_type_parameters_test.ts
filmaj: Done in 169620e

shapirone: Oh actually for this one, I think we should keep the test but instead of using `DefineProperty`, we should just define the Custom Type directly

```ts
  const myType = DefineType({
    name: "custom",
    type: Schema.types.object,
    properties: {
      required_property: { type: "string" },
      optional_property: { type: "string" },
    },
    required: ["required_property"],
  });
```

filmaj: Yeah I think the test on line 137 in this file has that covered, but if not happy to add another one.

tests/integration/functions/runtime_context/string_parameters_test.ts
filmaj: Done in 9ca9bff

tests/integration/functions/runtime_context/string_parameters_test.ts
filmaj: Done in 09eca8a

shapirone (approved):
Thank you for doing this!

PR #136, opened 2023-02-07
Built in function gen improve

WilliamBergamin:
###  Summary

This is a draft PR used to get feedback on approaches.

In this PR the types were improved, Im looking for feedback on wether the `functions` payload mapped to `FunctionsPayload` type should be mapped to the `ManifestFunctionSchema` type and then translated to proper typescript files

In short `payload` -> `FunctionsPayload`  -> `ManifestFunctionSchema` -> `typescript string`

Alternatively the `FunctionsPayload` would be directly mapped to the `typescript string`. The advantage I saw in using the `ManifestFunctionSchema` is that the `typescript string` creates a file that defines a `ManifestFunctionSchema`. I am still unsure if this additional complexity will benefit us in the long run.

alternative: `payload` -> `FunctionsPayload` -> `typescript string`


filmaj:
Is it correct to assume that the various generated function and test files in this PR come from running the new generation script?

src/schema/slack/functions/_scripts/src/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    paramDef.required = param.required;
+  }
+  return paramDef as ParameterDefinition;
+};
+
+const getManifestFunctionParameters = (
+  functionParameters: FunctionParameter[],
+): ManifestFunctionParameters => {
+  return {
+    properties: Object.fromEntries<ParameterDefinition>(
```
filmaj, line 83: TIL about `fromEntries`, nice!

src/schema/slack/functions/_scripts/src/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    },
+    [],
+  );
+}
+
+const getParameterDefinition = (
+  param: FunctionProperty,
+): ParameterDefinition => {
+  // deno-lint-ignore no-explicit-any
+  const paramDef: { [key: string]: any } = {
```
filmaj, line 58: Not required, but FYI there is also the typescript utility type [`Record<KeyType, ValueType>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) that you can use instead of `{ [key: string}: any }`, e.g. `Record<string, any>`.

src/schema/slack/functions/_scripts/src/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    type: param.type,
+    description: param.description,
+  };
+  if (param.description) {
+    paramDef.description = param.description;
```
filmaj, line 63: Isn't this already duplicated in line 60?

filmaj (commented):
The `ManifestFunctionSchema` approach looks fine to me so far!

WilliamBergamin:
> Is it correct to assume that the various generated function and test files in this PR come from running the new generation script?

Yess exactly, still not sure if we want to include the changes in this PR or not, I might ask for the original `function.json` file in order to only generate the test of the existing functions

src/schema/slack/functions/_scripts/src/utils.ts
WilliamBergamin: Yes this seems to be a duplicate, but this brings me to an issue of this current approach

This field check must be done twice once here and then again when the function [gets written to typescript ](https://github.com/WilliamBergamin/deno-slack-sdk/blob/built-in-function-gen-improve/src/schema/slack/functions/_scripts/src/templates/template_function.ts#L47) this leads to code duplication

filmaj:
> order to only generate the test of the existing functions

I think this makes sense, also easier to review this PR as it is only focussed on the generation (and not on the new additions)

src/schema/slack/functions/_scripts/src/templates/template_mod.ts
```diff
@@ -0,0 +1,37 @@
...
+  autogeneratedComment,
+  getFunctionName,
+  renderFunctionImport,
+} from "./template_utils.ts";
+import { FunctionRecord } from "../types.ts";
+
+const renderFunctionPropWithType = (defineFunctionInput: FunctionRecord) =>
+  `${getFunctionName(defineFunctionInput.callback_id)}: ${
+    getFunctionName(defineFunctionInput.callback_id)
+  }`;
```
shapirone, line 11: nit: I think we could get away with just a single call here using [Object initializer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer)
 
```suggestion
const renderFunctionPropWithType = (defineFunctionInput: FunctionRecord) =>
  `${getFunctionName(defineFunctionInput.callback_id)}`;
```

src/schema/slack/functions/_scripts/src/templates/template_utils.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function getParameterType(type: AllowedTypeValue): string {
+  return type instanceof Object ? type.id : type;
+}
+
+/** @description returns a list of all fn parameters */
+const getParameterList = (fn: FunctionRecord): FunctionParameter[] => [
+  ...fn?.input_parameters ?? [],
+  ...fn?.output_parameters ?? [],
```
shapirone, line 34: nit: We don't need the optional chain here do we?? `fn` is always going to be defined

src/schema/slack/functions/_scripts/src/templates/test/template_utils_test.ts
```diff
@@ -0,0 +1,94 @@
+export { assertEquals } from "https://deno.land/std@0.152.0/testing/asserts.ts";
```
shapirone, line 1: I don't think we should be exporting this, just import from `dev_deps.ts`

src/schema/slack/functions/_scripts/src/templates/test/template_utils_test.ts
```diff
@@ -0,0 +1,94 @@
...
+  autogeneratedComment,
+  getFunctionName,
+  getSlackCallbackId,
+  renderFunctionImport,
+  renderTypeImports,
+} from "../template_utils.ts";
+import {
+  assertEquals,
+  assertStringIncludes,
+} from "https://deno.land/std@0.152.0/testing/asserts.ts";
```
shapirone, line 12: The [Deno standard for managing dependencies](https://deno.land/manual@v1.30.3/examples/manage_dependencies) is to use a `dev_deps` file to manage importing from external libraries for dev only dependencies

src/schema/slack/functions/_scripts/src/templates/test/template_utils_test.ts
```diff
@@ -0,0 +1,94 @@
...
+          type: "integer",
+        },
+      },
+    ],
+  };
+  const actual = renderTypeImports(dfi);
+  assertStringIncludes(actual, "SchemaTypes");
+  assertStringIncludes(actual, "SlackTypes");
+  assertStringIncludes(actual, "InternalSlackTypes");
+});
```
shapirone, line 94: Loving all these script tests, thanks so much for adding them

src/schema/slack/functions/_scripts/src/templates/test_template.ts
```diff
@@ -0,0 +1,137 @@
...
+  typescript.push(renderTypeImports(defineFunctionInput));
+  typescript.push(renderFunctionImport(defineFunctionInput.callback_id));
+  typescript.push("");
+  typescript.push(
+    `Deno.test("${
+      getFunctionName(defineFunctionInput.callback_id)
+    } generates valid FunctionManifest", ${
+      renderFunctionManifestTest(defineFunctionInput)
+    });`,
+  );
```
shapirone, line 124: Ya know, while I don't find testing this for built-ins to be particularly valuable, I'm very impressed at how simple the implementation is for it — well done!

src/schema/slack/functions/_scripts/src/templates/test_template.ts
```diff
@@ -0,0 +1,137 @@
...
+    case SchemaTypes.number:
+      return `${name}: 1.5`;
+    case SchemaTypes.array:
+      return `${name}: []`;
+    case SchemaTypes.object:
+      return `${name}: {}`; // TODO make this recursive
+    default:
+      return `${name}: "test"`;
+  }
+};
```
shapirone, line 66: Honestly, our typing on step inputs is very liberal and we intentionally always need to support them accepting a string since step inputs can be a templated variable reference that looks something like `"{{inputs.my_channel}}"`

Because of this, we could probably just assign them all to a string and get rid of this switch case.

src/schema/slack/functions/_scripts/src/templates/test_template.ts
```diff
@@ -0,0 +1,137 @@
...
+  typescript.push(`const actual = testWorkflow.steps[0].export();`);
+  typescript.push("");
+  typescript.push(
+    `assertEquals(actual.function_id, "${getSlackCallbackId(dfi)}");`,
+  );
+  typescript.push(`assertEquals(actual.inputs, ${inputParameters});`);
+  return `() => {${typescript.join("\n")}}`;
+};
+
+export function SlackTestFunctionTemplate(
```
shapirone, line 103: My opinion is that the most valuable things we can test for built-ins is
- the optionality for their step inputs (making sure that all required fields must be set, making sure non-required field can be set but are dont have to be)
- that we can access all outputs regardless of if they are required or not

There are lots of combinations we could generate tests for in this case, but here's an idea of what this could look like for `ArchiveChannel`
```ts
/** testing inputs */
const step = testWorkflow.addStep(ArchiveChannel, { channel_id: "test" });
// @ts-expect-error not providing `channel_id` leads to a type error
testWorkflow.addStep(ArchiveChannel, {});

/** testing outputs */
assert<CannotBeUndefined<typeof step.outputs.channel_id>>(true);
```

shapirone (commented):
Really well done William! This didn't feel like a Draft PR at all 😁

src/schema/slack/functions/_scripts/src/write_function_files.ts
```diff
@@ -1,31 +1,39 @@
...
+
+  console.log(
+    `Generating unit test for Slack Function: ${
+      greenText(functionRecord.callback_id)
+    }`,
+  );
+  const templateTestString = SlackTestFunctionTemplate(functionRecord);
+  const testFilename = `../${functionRecord.callback_id}_test.ts`;
+
+  await Deno.writeTextFile(testFilename, templateTestString);
```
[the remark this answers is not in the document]
WilliamBergamin: Good catch I did not notice this!

I've opted to `await` until the files are generated before executing the `console.log` statement

src/schema/slack/functions/_scripts/src/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+const reset = "\x1b[0m";
+
+export const greenText = (text: string) => green + text + reset;
+export const yellowText = (text: string) => yellow + text + reset;
+export const redText = (text: string) => red + text + reset;
+
+export function isValidFunctionFile(fileName: string) {
+  return !(fileName.endsWith("_test.ts") ||
+    DIRECTORY_PATH_ALLOW_LIST.includes(fileName));
+}
```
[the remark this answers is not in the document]
WilliamBergamin: Good catch 💯  this is leftovers from the deletion script

src/schema/slack/functions/_scripts/src/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function loadFunctionsJson(): Promise<FunctionsPayload> {
+  return await Deno.readTextFile(
+    FUNCTIONS_JSON_PATH,
+  ).then(JSON.parse);
+}
+
+export function getSlackFunctions(functionsPayload: FunctionsPayload) {
+  return functionsPayload.functions.filter((fn) => !fn.app_id);
+}
```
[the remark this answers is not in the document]
WilliamBergamin: yess, I think I saw something about `app_id` that could break this script so using `"type": "builtin"` should fix it

src/schema/slack/functions/_scripts/src/templates/template_function.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (property.description) {
+    typescript.push(`description: "${property.description}"`);
+  }
+  if (property.title) {
+    typescript.push(`title: "${property.title}"`);
+  }
+  if ("items" in property) {
+    typescript.push(`items: ${propertyToTypeScript(property.items)}`);
+  }
+  if ("properties" in property) {
```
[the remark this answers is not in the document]
WilliamBergamin: Here this does not work because my property is of type `FunctionProperty` and I'm [using narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) (Fil showed me this yesterday) to access the fields

src/schema/slack/functions/_scripts/src/templates/template_mod.ts
WilliamBergamin: Noice, I felt like this function was ugly I'm happy to remove it

src/schema/slack/functions/_scripts/src/templates/template_utils.ts
WilliamBergamin: Yes, same with `output_parameters` and `input_parameters` because of how I have defined the types

src/schema/slack/functions/_scripts/src/templates/test_template.ts
WilliamBergamin: I had already written the code for the test, figured It did not hurt to leave it in 💡

src/schema/slack/functions/_scripts/src/templates/test_template.ts
WilliamBergamin: I've implemented a new generated test to cover outputs 💯

PR #140, opened 2023-02-08
Built in function gen improve

WilliamBergamin:
###  Summary

This PR aims to improve the code generation script the main improvement adds generated unit tests for built in functions.

JIRA: HERMES-3364

A lot of these changes are caused by the change in generation method, the manual changes for this PR are found in the `_scripts` folder

Testing steps:
1. pull the repo
2. navigate to `src/schema/slack/functions/_scripts`
3. Use this tester to fetch the `FunctionPayload` https://api.slack.com/methods/functions.list/test
4. Follow & review the README.md (`src/schema/slack/functions/_scripts/README.md`)


filmaj (commented):
The code looks pretty good to me! But! I got errors running it: 

```
➜ ./generate
Cleaning folder directory
Generating code & tests for Slack Function: add_channel_list_to_section
Generating code & tests for Slack Function: add_channel_manager
Generating code & tests for Slack Function: add_channel_to_channel_section
Generating code & tests for Slack Function: add_pin
Generating code & tests for Slack Function: add_user_to_usergroup
Generating code & tests for Slack Function: archive_channel
Generating code & tests for Slack Function: channel start message
Generating code & tests for Slack Function: create_channel
Generating code & tests for Slack Function: create_channel_section
Generating code & tests for Slack Function: create_usergroup
Generating code & tests for Slack Function: delay
Generating code & tests for Slack Function: find_channels_by_keywords
Generating code & tests for Slack Function: invite_user_to_channel
Generating code & tests for Slack Function: open_form
Generating code & tests for Slack Function: remove_user_from_usergroup
Generating code & tests for Slack Function: reply_in_thread
Generating code & tests for Slack Function: select_channels
Generating code & tests for Slack Function: send_dm
Generating code & tests for Slack Function: send_ephemeral_message
Generating code & tests for Slack Function: send_message
Generating code & tests for Slack Function: update_channel_topic
Generated 21 built-in functions with their unit tests
Updated functions module export
Formatting Slack function files...
Error formatting: /Users/fmaj/src/deno-slack-sdk/src/schema/slack/functions/add_channel_manager.ts
   Expected ',', got ':' at /Users/fmaj/src/deno-slack-sdk/src/schema/slack/functions/add_channel_manager.ts:16:12

  items: type: SlackTypes.user_id}},
             ~
Error formatting: /Users/fmaj/src/deno-slack-sdk/src/schema/slack/functions/select_channels_test.ts
   Expected ',', got ':' at /Users/fmaj/src/deno-slack-sdk/src/schema/slack/functions/select_channels_test.ts:15:12

  items: type: SlackTypes.channel_id}},
             ~
```

src/schema/slack/functions/_scripts/src/deps.ts
```diff
@@ -0,0 +1 @@
+export { pascalCase } from "https://deno.land/x/case@v2.1.0/mod.ts";
```
shapirone, line 1: Two questions
- Is there a reason we are giving this a separate `deps.ts` file instead of using our existing `src/deps.ts`?
- Should this be in a `dev_deps.ts` file since we only use it in this generation script?

src/schema/slack/functions/_scripts/src/templates/template_function.ts
```diff
@@ -0,0 +1,149 @@
...
+    typescript.push(`description: "${property.description}"`);
+  }
+  if (property.title) {
+    typescript.push(`title: "${property.title}"`);
+  }
+  if ("items" in property) {
+    typescript.push(`items: ${propertyToTypeScript(property.items)}`);
+  }
+  if ("properties" in property) {
+    typescript.push(`additionalProperties: ${property.additionalProperties}`);
```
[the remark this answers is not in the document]
shapirone: The property in this case is a `FunctionProperty` that is an object, so because it comes from how the backend stores functions there may be a shape difference between what we require — tbh I don't know if the backend fills `additionalProperties` and `required` when it stores them. Safe bet is to handle the undefined case!

src/schema/slack/functions/_scripts/src/test/data/function.json
```diff
@@ -0,0 +1,115 @@
+{
```
shapirone, line 1: Probably a good idea to add an array case here based on Fil finding an error when arrays come back in the payload

src/schema/slack/functions/_scripts/src/deps.ts
WilliamBergamin: I was unsure where to put this, should have asked in the PR

Thanks for clearing it up 💯  I think `dev_deps.ts` is the best place to put this since it won't be required at runtime

src/schema/slack/functions/_scripts/src/templates/template_function.ts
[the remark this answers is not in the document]
WilliamBergamin: Good catch!!
I was also unsure about what was required and what wasn't, will fix to deal with worst case

src/schema/slack/functions/_scripts/src/test/data/function.json
WilliamBergamin: @shapirone I just ran some more tests here and it seems typescript does not support Type Safe Json parsing, this test only validates that the `builtin` parsing work properly. TIL

I still added the array type in the file regardless. Not sure if we want to do more unit test surrounding this

WilliamBergamin: Do you think this is worth implementing [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) and assuring this is type safe?

src/schema/slack/functions/_scripts/src/templates/template_function.ts
```diff
@@ -0,0 +1,153 @@
...
+  if ("properties" in property) {
+    typescript.push(
+      `properties: ${propertiesToTypeScript(property.properties)}`,
+    );
+    if (property.additionalProperties) {
+      typescript.push(`additionalProperties: ${property.additionalProperties}`);
+    }
+    if (property.required) {
+      typescript.push(`required: ${JSON.stringify(property.required)}`);
+    }
```
shapirone, line 61: nit: It'd be nice to still set these, regardless if they're on the property
```suggestion
    typescript.push(
      `additionalProperties: ${property.additionalProperties ?? true}`,
    );
    typescript.push(`required: ${JSON.stringify(property.required ?? [])}`);
```

src/schema/slack/functions/_scripts/src/test/data/function.json
shapirone: Ah I didn't realize that this file was just for the `builtin` filtering unit test, I was thinking we'd be able to use this to test how an array would come out. Might want to have a separate unit test for `propertyToTypescript` and test the various cases.

Where was type safety coming into play here?

src/schema/slack/functions/_scripts/src/templates/template_function.ts
```diff
@@ -0,0 +1,153 @@
...
+  );
+  if (property.description) {
+    typescript.push(`description: "${property.description}"`);
+  }
+  if (property.title) {
+    typescript.push(`title: "${property.title}"`);
+  }
+  if ("items" in property) {
+    typescript.push(`items: ${propertyToTypeScript(property.items)}`);
+  }
```
shapirone, line 51: Array items can't recurse directly, they need to be wrapped in an object

```suggestion
  if ("items" in property) {
    typescript.push(`items: {${propertyToTypeScript(property.items)}}`);
  }
```

src/schema/slack/functions/_scripts/src/test/data/function.json
WilliamBergamin: `SlackFunctionTemplate` `SlackFunctionModTemplate` and `SlackTestFunctionTemplate` rely on the fact that they expect a parameter of type `FunctionRecord`, but since `JSON.parse` does not actually do any type safety assertions  and simply casts the object to `FunctionRecord` all the type assumptions made may be wrong

The contents of `function.json` may be rubbish but `JSON.parse` would still cast it to `FunctionRecord`

src/schema/slack/functions/_scripts/src/templates/template_function.ts
WilliamBergamin: Good catch 💯  I made this more generic to `properties`

src/schema/slack/functions/_scripts/src/templates/template_function.ts
WilliamBergamin: I've implemented this but because of type narrowing I had to do this

```js
if ("additionalProperties" in property && property.additionalProperties) {
  typescript.push(`additionalProperties: ${property.additionalProperties}`);
}
```

src/schema/slack/functions/_scripts/src/templates/template_function.ts
WilliamBergamin: I resolved this by wrapping the property object with `{  }` on L63

src/schema/slack/functions/_scripts/src/test/data/function.json
WilliamBergamin: `deno.lint` will fail if the generated file is invalid, this is the solution we agreed to 🧠

src/schema/slack/functions/_scripts/src/templates/template_function.ts
shapirone: Sorry, I think maybe the commit suggestion feature is causing some confusion here. What we want is that an `ObjectFunctionProperty` will always set it's `additionalProperties` and `required` keys to their defaults, regardless of if they are `undefined` (since we are saying they're optional). This should be totally valid if it's done within the type guard that guarantees it's an object (in this case `if ("properties" in property)`).

```ts
  if ("properties" in property) {
    // Inside the type guard, we know that `property` is `ObjectFunctionProperty`
    // so we can access keys that exist on that type
    typescript.push(
      `properties: ${propertiesToTypeScript(property.properties)}`,
    );
    typescript.push(
      `additionalProperties: ${property.additionalProperties ?? true}`, // if `undefined`, set to `true`
    );
    typescript.push(`required: ${JSON.stringify(property.required ?? [])}`); // if `undefined`, set to an empty array
  }
```

The latest change not only removes the narrowed type, but it also won't set `additionalProperties` or `required` if they're `undefined`.

src/schema/slack/functions/_scripts/src/write_function_files.ts
```diff
@@ -1,31 +1,47 @@
...
+      }`,
+    );
+    if (!VALID_FILENAME_REGEX.test(functionRecord.callback_id)) {
+      console.log(
+        `${
+          redText(
+            "Unable to generate files",
+          )
+        }, invalid characters in callback_id: "${functionRecord.callback_id}"`,
+      );
```
shapirone, line 28: Oh this is nice! What do you think about
```ts
console.log(
  `${redText("FAILURE:")} Invalid characters in callback_id: ${
    redText(functionRecord.callback_id)
  }`,
);
```

to produce something like
image

WilliamBergamin: Ohh yeah thats much better 💯  will push this now

WilliamBergamin: Done 🟢

src/schema/slack/functions/_scripts/src/templates/template_function.ts
```diff
@@ -0,0 +1,153 @@
...
+    typescript.push(
+      `properties: ${propertiesToTypeScript(property.properties)}`,
+    );
+  }
+  if ("additionalProperties" in property && property.additionalProperties) {
+    typescript.push(`additionalProperties: ${property.additionalProperties}`);
+  }
+  if ("required" in property && property.required) {
+    typescript.push(`required: ${JSON.stringify(property.required)}`);
+  }
```
shapirone, line 62: Sorry, I think maybe the commit suggestion feature is causing some confusion here. What we want is that an ObjectFunctionProperty will always set it's additionalProperties and required keys to their defaults, regardless of if they are undefined (since we are saying they're optional). This should be totally valid if it's done within the type guard that guarantees it's an object (in this case if ("properties" in property)).

```ts
  if ("properties" in property) {
    // Inside the type guard, we know that `property` is `ObjectFunctionProperty`
    // so we can access keys that exist on that type
    typescript.push(
      `properties: ${propertiesToTypeScript(property.properties)}`,
    );
    typescript.push(
      `additionalProperties: ${property.additionalProperties ?? true}`, // if `undefined`, set to `true`
    );
    typescript.push(`required: ${JSON.stringify(property.required ?? [])}`); // if `undefined`, set to an empty array
  }
```

The latest change not only removes the narrowed type, but it also won't set additionalProperties or required if they're undefined.

shapirone (changes requested):
I think I put a comment on a resolved thread 🤦

src/schema/slack/functions/_scripts/src/templates/template_function.ts
WilliamBergamin: Ohh I see, this is cleaner code as well, I think I miss-understood the original comment

WilliamBergamin: Done 🟢

WilliamBergamin: I opted to implement type predicates instead of writing comments

shapirone (approved):
Looks great to me! Huge improvement, thanks for taking care of this 🙇

PR #143, opened 2023-03-01
Integrate a better CLI utility for generating code coverage reports.

filmaj:
###  Summary

Small tweak to the coverage task, I think the output is much nicer and it also gives you a project-wide total (which the built-in deno CLI `deno coverage` tool does not):

Screenshot 2023-03-01 at 12 45 50 PM


shapirone (approved):
👍

PR #145, opened 2023-03-03
Ignore empty array values for manifest's `externalAuthProviders`

filmaj:
###  Summary

Re: HERMES-4933

### DIFF
--- a/src/manifest/mod.ts
+++ b/src/manifest/mod.ts
@@ -287,8 +287,8 @@ export class SlackManifest {
 
     // External Auth providers
     if (
-      def.externalAuthProviders && Array.isArray(def.externalAuthProviders) &&
-      def.externalAuthProviders.length > 0
+      def.externalAuthProviders &&
+      def.externalAuthProviders?.length > 0
     ) {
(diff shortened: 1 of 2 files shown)


src/manifest/mod.ts
```diff
@@ -286,7 +286,10 @@ export class SlackManifest {
     }
 
     // External Auth providers
-    if (def.externalAuthProviders) {
+    if (
+      def.externalAuthProviders && Array.isArray(def.externalAuthProviders) &&
+      def.externalAuthProviders.length > 0
+    ) {
```
shapirone, line 292:
```ts
if (def.externalAuthProviders?.length) { ... }
```

Could we just get away with the above? TS should ensure that the value passed is an array, then if `def.externalAuthProviders` doesn't exist or has a length of 0 it'll be falsy

filmaj: deno-ts still complains about it being possibly undefined but I can bring it down to:

```
if (
      def.externalAuthProviders &&
      def.externalAuthProviders?.length > 0
    ) {
```

shapirone (approved):
Ezpz, LGTM!

PR #147, opened 2023-03-14
Surface custom type prototype

WilliamBergamin:
###  Summary

This PR aims to introduce a tool that allows users to surface usable types from the returned objects of `DefineType`, `DefineProperty` and `DefineFunction`

2 utilities are now available to the users 
`FunctionRuntimeType`
`RuntimeType`

Example of usage:
```typescript
  const functionDefinition = DefineFunction({
    callback_id: "test_function",
    title: "Test function",
    source_file: "functions/test_function.ts",
    input_parameters: {
      properties: {
        channel_id: {
          type: Schema.slack.types.channel_id,
        },
      },
      required: ["channel_id"],
    },
    output_parameters: {
      properties: {
        is_valid: {
          type: Schema.types.boolean,
        },
      },
      required: ["is_valid"],
    },
  });

  type FunctionDefinitionType = FunctionRuntimeType<typeof functionDefinition>;

  type FunctionDefinitionInputType = FunctionDefinitionType["args"]["inputs"];
  type FunctionDefinitionOutputType = FunctionDefinitionType["outputs"];

  const printInputs = (input: FunctionDefinitionInputType) => {
    console.log(input.channel_id);
  };

  const printOutputs = (output: FunctionDefinitionOutputType) => {
    console.log(output.is_valid);
  };
```

Looking for feedback on:
* Solution meets expectations
* Usability
* Test cover the proper use cases and are placed in the correct location

### Requirements


filmaj (commented):
I like the direction this is going in!

Left one comment about re-using the `FunctionContext` type for the function runtime helper type you added - just to try to not copy/paste the `{ inputs, outputs }` shape around too much and re-use what we can.

src/functions/types.ts
```diff
@@ -396,3 +397,14 @@ export type RuntimeUnhandledEventContext<
     // deno-lint-ignore no-explicit-any
     body: any;
   };
+
+export type DefineFunctionReturnType = ReturnType<typeof DefineFunction>;
+
+export type FunctionRuntimeType<
+  Function extends DefineFunctionReturnType,
+> = Function["definition"] extends
+  FunctionDefinitionArgs<infer I, infer O, infer RI, infer RO> ? {
```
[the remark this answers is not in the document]
WilliamBergamin: Thanks for pointing this out, my latest commit introduces this 💯

src/functions/types.ts
```diff
@@ -396,3 +397,32 @@ export type RuntimeUnhandledEventContext<
     // deno-lint-ignore no-explicit-any
     body: any;
   };
+
+type DefineFunctionReturnType = ReturnType<typeof DefineFunction>;
```
filmaj, line 401: Since this is only used in one other place in this file, I would remove it. I don't think it offers much additional value, and we already suffer from having too many types and type aliases in this project as it is.

My suggestion is to just use `ReturnType<typeof DefineFunction>` directly where you need it

src/functions/types.ts
```diff
@@ -396,3 +397,32 @@ export type RuntimeUnhandledEventContext<
...
+
+type FunctionRuntimeReturnArgs<
+  FunctionReturnType extends
+    | FunctionHandlerReturnArgs<UnknownRuntimeType>
```
filmaj, line 405: Just to confirm / my own education, use of `UnknownRuntimeType` here is only to avoid the `any` warnings from deno, yes?

src/functions/types.ts
```diff
@@ -396,3 +397,32 @@ export type RuntimeUnhandledEventContext<
     // deno-lint-ignore no-explicit-any
     body: any;
   };
+
+type DefineFunctionReturnType = ReturnType<typeof DefineFunction>;
+
+type FunctionRuntimeReturnArgs<
```
filmaj, line 403: Can we add a comment above here to summarize what this type is for? If I understand correctly, it simply it just unwraps the return from a function in case it is wrapped in a promise.

tests/integration/functions/types_test.ts
```diff
@@ -0,0 +1,162 @@
...
+          callback_id: "",
+          title: "",
+        },
+        function_execution_id: "",
+        type: "function_executed",
+      },
+    },
+  };
+
+  assert<CanBe<typeof expected, Actual>>(true);
```
filmaj, line 119: I think this assertion will always be true, since we manually assign the `Actual` type to the `expected` variable.

You're on the right track, though, for how to test this.. hmm.. maybe one thing we can try instead is to actually implement the `testFunctionDefinition` runtime logic / handler, and then fake 'executing' it to retrieve the output of the function, then compare the type of the function handler output to the type returned by the `FunctionRuntimeType` type?

For an example of this, check out the [output parameter integration tests](https://github.com/slackapi/deno-slack-sdk/blob/main/tests/integration/functions/runtime_context/output_parameters_test.ts):

- [here](https://github.com/slackapi/deno-slack-sdk/blob/main/tests/integration/functions/runtime_context/output_parameters_test.ts#L25-L32) we implement a function handler that meets the function definition in the test. You could do a similar thing in this test (and move a majority of what the `expected` variable you have in this test sets to inside the function handler). The nice thing about the various `*SlackFunctionHandler` SDK types is that they force the dev writing the handler to return the required output parameters, as per their function definition.
- [here](https://github.com/slackapi/deno-slack-sdk/blob/main/tests/integration/functions/runtime_context/output_parameters_test.ts#L33-L34) we use some of our SDK integration testing utilities to create a function runtime 'context' object (the thing that has all the extras like token, env vars, team_id, etc.) useful for testing, and feed it to the handler we implemented on line 25.
- with the above two steps, we can then retrieve the output from the function! Once you have the actual runtime output from the function, you can use the exact kind of `assert` you use here on it to make sure the types are compatible.

Hit me up if you wanna pair on this, happy to help!

tests/integration/functions/types_test.ts
```diff
@@ -0,0 +1,162 @@
...
+    },
+    output_parameters: {
+      properties: {},
+      required: [],
+    },
+  });
+
+  type Actual = FunctionRuntimeType<typeof testFunctionDefinition>;
+
+  const expected: Actual = {
```
filmaj, line 139: Same comment as above - I'm not sure this test will ever fail if we manually assign the types we are comparing.

tests/integration/functions/types_test.ts
```diff
@@ -0,0 +1,162 @@
...
+          callback_id: "",
+          title: "",
+        },
+        function_execution_id: "",
+        type: "function_executed",
+      },
+    },
+  };
+
+  assert<CanBe<typeof expected, Actual>>(true);
```
filmaj, line 161: Careful with the order of generics/type parameters to `CanBe`! `CanBe<T, undefined>` is much different than `CanBe<undefined, T>`! I think about this as set membership: `CanBe<T, U>` translates to "is `U` included in the set of `T`?", or also "can `T` be a `U`?" The implication being that, if the assertion is true, we expect that T is a superset of U.

Possibly we may want to clarify this at the `CanBe` level under `src/test_utils.ts` via documentation and/or better naming of the type parameters; `T` and `U` are pretty generic and unclear. Or maybe even renaming the utility type to make it clear! Open to suggestions.

tests/integration/types_test.ts
```diff
@@ -0,0 +1,303 @@
...
+  const customType = DefineType({
+    title: "Title",
+    description: "Description",
+    name: "Name",
+    type: Schema.types.boolean,
+  });
+
+  type Actual = RuntimeType<typeof customType>;
+
+  assert<CanBe<boolean, Actual>>(true);
```
filmaj, line 16: Nice! Good test, but one small change I would do here is, at least for this specific test case, I would replace `CanBe` with `IsExact`. `CanBe` tests set membership, whereas `IsExact` tests set equality. In this case, since the custom type.. type is `boolean`, we expect its runtime type to only be `boolean` - nothing else. So we want to test set equality.

I think that applies to most of the assertions in the tests you wrote.

For example, consider the following types, and see how `CanBe` vs. `IsExact` differ:

```typescript
type bigSet = boolean | string;
type smallSet = boolean;
assert<CanBe<boolean, Actual>>(true);
```

tests/integration/types_test.ts
```diff
@@ -0,0 +1,303 @@
...
+    description: "Description",
+    name: "Name",
+    type: Schema.types.object,
+    properties: {},
+    required: [],
+  });
+
+  type Actual = RuntimeType<typeof customType>;
+
+  const expected = {};
```
filmaj, line 125: I'm not sure what the implied type of an empty object is - this one might be tricky!

But, I think most of your tests will require feeding these custom types / defined properties through a function definition, which then we have to pump through the test utilities we have to create a function handler, invoke the handler, and look at how the input and/or output types compare.

filmaj (changes requested):
Looking good!

I left a bunch of comments, mostly around the tests. The general theme I see in this latest round is:

- we need to be careful about what we are asserting in the tests, in particular, the subtle differences between `CanBe` and `IsExact` type assertion helpers.
- I think we'll need to implement the full function runtime handlers in each of the tests in order to pull out a) the return/output types from the function but also b) assert on the input parameter types from within the function handler.

Good job though! This stuff is pretty gnarly and tough to wrap your head around, at least for me it is, and you're doing great. Well done 👍

src/types.ts
```diff
@@ -10,6 +16,31 @@ export type {
...
+
+// deno-lint-ignore no-explicit-any
+type AnyCustomType = CustomType<any, any, any>;
+
+type AnyTypedObjectParameterDefinition = TypedObjectParameterDefinition<
+  // deno-lint-ignore no-explicit-any
+  any,
+  // deno-lint-ignore no-explicit-any
+  any
+>;
```
shapirone, line 31: Can we export these from the relevant directories? That'll help us keep this `src/types.ts` file clean for things we want to surface as our recommended API for devs

src/types.ts
```diff
@@ -10,6 +16,31 @@ export type {
   RuntimeSlackFunctionHandler as SlackFunctionHandler,
 } from "./functions/types.ts";
 
+// ----------------------------------------------------------------------------
+// Runtime Types
+// ----------------------------------------------------------------------------
+
+// deno-lint-ignore no-explicit-any
+type AnyCustomType = CustomType<any, any, any>;
```
shapirone, line 24: Rather than passing `any`, we can pass the exact shape that these expect. [We actually already do this for `TypedObjectParameterDefinition` and call it `TypedObjectParameter`](https://github.com/slackapi/deno-slack-sdk/blob/main/src/parameters/definition_types.ts#L98).

```ts
type AnyCustomType = CustomType<
  TypedObjectProperties,
  TypedObjectRequiredProperties<TypedObjectProperties>,
  CustomTypeDefinition<TypedObjectProperties, TypedObjectRequiredProperties<TypedObjectProperties>>
>;

type AnyTypedObjectParameterDefinition = TypedObjectParameter
```

src/functions/types.ts
shapirone: Similar to my comment above, we could directly use the expected shape instead.

```ts
FunctionHandlerReturnArgs<FunctionParameters>
```

src/types.ts
```diff
@@ -10,6 +16,31 @@ export type {
...
+/**
+ * @description Used to surface runtime Typescript types from Defined Types and Properties
+ */
+export type RuntimeType<
+  T extends
+    | AnyCustomType
+    | AnyTypedObjectParameterDefinition,
+> = T extends AnyCustomType ? FunctionInputRuntimeType<T["definition"]>
+  : T extends AnyTypedObjectParameterDefinition ? FunctionInputRuntimeType<T>
+  : never;
```
shapirone, line 42: Do we need the catch-all else case to be never, or could we just have the catch-all directly use the parameter definition? If we do the latter, we don't need to introduce more cases if we decide to support other parameter types.

For instance, I foresee this could eventually become

```ts
export type RuntimeType<
  T extends
    | AnyCustomType
    | AnyTypedObjectParameterDefinition
    | ParameterDefinition
> = T extends AnyCustomType ? FunctionInputRuntimeType<T["definition"]>
  : FunctionInputRuntimeType<T>;
```

src/functions/types.ts
WilliamBergamin: Yes exactly this simply unwraps the promise, added a comment 💯

src/functions/types.ts
WilliamBergamin: This makes sense, Ive implemented this 👍

src/functions/types.ts
WilliamBergamin: I've improved the types, this isn't `AnyScript`

tests/integration/functions/types_test.ts
WilliamBergamin: Thanks for the information 💯  I've implemented this and it seems to work

tests/integration/types_test.ts
WilliamBergamin: Ive updated every test to use `IsExact` where possible 🥇

tests/integration/types_test.ts
WilliamBergamin: Replaced `{}` with `Record<string | number | symbol, never>` where ever possible

src/functions/types.ts
filmaj: TBH I love anyscript

src/types.ts
WilliamBergamin: Tried iterating on this idea but I was not able to make it work, some OAuth types seem to be creating conflicts

src/types.ts
WilliamBergamin: Since `AnyCustomType` `AnyTypedObjectParameterDefinition` are replaced with `ICustomType` and `TypedObjectParameter` I don't think this will be needed

src/types.ts
shapirone: :chefs-kiss:

src/types.ts
WilliamBergamin: I've generalized `RuntimeType` to accept any `ParameterDefinition` not only the `TypedObjectParameter` this seems to work as expected

tests/integration/functions/types_test.ts
```diff
@@ -0,0 +1,199 @@
...
+    num: 0.1,
+    string: "",
+    arr: [true],
+    obj: {
+      bool: true,
+    },
+  };
+
+  const validHandler: EnrichedSlackFunctionHandler<typeof TestFn.definition> =
+    () => {
```
filmaj, line 94: One thing you might want to assert on here is that the type of the arguments provided to the function handler match the `Actual.args` type.

```suggestion
    (args) => {
      assert<IsExact<typeof args, Actual["args"]>>(true);
```

I think this is a slightly better testing approach than testing the type of what is returned by `createContext` below, because `createContext` is a testing utility we wrote, that might have some shortcuts or use other code paths that are not user/developer-facing. Therefore, asserting on the output of `createContext` tests our own test utilities rather than testing the parts that app devs interface with.

filmaj (changes requested):
Nice! Left one comment, nicely done, I think this is pretty close!

src/types.ts
shapirone: Ahhhh I see! Thank you for looking at this. We've found before that the else case of a conditional type does not narrow the available type down based on the check, which means that when we hit `: FunctionInputRuntimeType<T>`, TS still infers T as `T extends ICustomType | ParameterDefinition` rather than just `T extends ParameterDefinition`

src/functions/types.ts
```diff
@@ -396,3 +397,40 @@ export type RuntimeUnhandledEventContext<
...
+  >,
+> = {
+  args: Parameters<EnrichedFunction>[number];
+  outputs: FunctionRuntimeReturnArgs<ReturnType<EnrichedFunction>>["outputs"];
+};
+
+/**
+ * @description Used to surface function runtime typescript types from defined functions
+ */
+export type FunctionRuntimeType<
```
shapirone, line 432: Food for thought — I read `FunctionRuntimeType` as the type of the function runtime signature. I think we could do better to indicate it's a utility type and what it's doing
1. Pluralize to `FunctionRuntimeTypes` to indicate we're returning multiple types
2. Use a verb to indicate it's a utility that is taking an action rather than responding with the type representation. Suggest `ExtractFunctionRuntimeTypes`

shapirone (changes requested):
This is looking good and really close! We do need to make sure we export the Function type.

Otherwise I noticed that `outputs` is coming back as potentially `undefined` and I dropped some comments on naming that we may want to hash out in chat our in a quick sync.

src/functions/types.ts
WilliamBergamin: I will implement both of these points and apply point `2.` to `RuntimeType` -> `ExtractRuntimeType` because using a verb had crossed my mind in the past and for consistency purposes

src/functions/types.ts
```diff
@@ -396,3 +397,40 @@ export type RuntimeUnhandledEventContext<
     // deno-lint-ignore no-explicit-any
     body: any;
   };
+
+/**
+ * @description Used to extract returned args from Async and Sync function handlers
+ */
+type FunctionRuntimeReturnArgs<
```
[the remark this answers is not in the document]
WilliamBergamin: This type is simply used to isolate the logic required to extract `FunctionRuntimeReturnArgs` which is the returned type from `AsyncFunctionHandler` or `SyncFunctionHandler`
`AsyncFunctionHandler` => `Promise<FunctionHandlerReturnArgs<OutputParameters>>`
`SyncFunctionHandler` => `FunctionHandlerReturnArgs<OutputParameters>`

I've rename the type to `ExtractFunctionRuntimeReturnArgs` because it handles the exaction for Async and Sync function types

src/types.ts
```diff
@@ -10,6 +14,21 @@ export type {
...
 } from "./functions/types.ts";
 
+// ----------------------------------------------------------------------------
+// Runtime Types
+// ----------------------------------------------------------------------------
+
+/**
+ * @description Used to surface runtime Typescript types from Defined Types and Properties
+ */
+export type RuntimeType<
```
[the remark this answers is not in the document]
WilliamBergamin: This function accept `CustomType` and any type produced by `DefineProperty` (any `ParameterDefinition`) I think this includes all the "Types" a user can create I renamed it to `ExtractRuntimeType` for consistency, lmk what you think

tests/integration/functions/types_test.ts
WilliamBergamin: Good catch based on this I think I've been able to better implement the assertions of this file

filmaj (approved):
LGTM!

src/functions/types.ts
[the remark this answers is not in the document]
shapirone: I like that

WilliamBergamin:
After further discussion, this feature will need to wait before being released

PR #150, opened 2023-03-24
Do not auto-add datastore-related scopes if they are missing

filmaj:
Instead of automatically adding the datastore-related scopes to the app manifest, we now warn about the missing scopes.

This functionality will _only_ work with CLI v1.21 or newer and deno-slack-hooks v0.10.0 or newer. Using an older CLI with this PR will break pretty much all the CLI commands (as this relies on the new protocol support). Using a CLI v1.21 or newer but deno-slack-hooks prior to 0.10.0 will cause the warning not to be displayed (but will cause no breakage).

Related to [HERMES-4446](https://jira.tinyspeck.com/browse/HERMES-4446).


src/manifest/mod.ts
```diff
@@ -180,20 +180,30 @@ export class SlackManifest {
...
+   * @returns {string[]} The user-defined manifest scopes from `definition.botScopes`
+   */
   private ensureBotScopes(): string[] {
-    const includedScopes = this.definition.botScopes || [];
-
-    // Add datastore scopes if necessary
+    // Warn about missing datastore scopes if app includes datastores
     if (Object.keys(this.definition.datastores ?? {}).length > 0) {
+      const missingScopes: string[] = [];
       const datastoreScopes = ["datastore:read", "datastore:write"];
```
WilliamBergamin, line 191: I know this is not part of the changes but should this be placed is some sort of a global constant location instead of  directly hard coded in the function?

filmaj: Yeah I thought about that too, but then it is only used here in this one place, and the performance impact of this is negligible, so I figured it would make the code more readable to keep the array near where it is used. Open to changing it if we think that is worth it.

shapirone (approved):
LGTM, comment on Slack about potentially updating the messaging in the future

PR #154, opened 2023-03-30
Remove deprecated built-in function parameters

shapirone:
###  Summary

This PR is to prep for our breaking changes release. I re-ran the generation script and manually removed the deprecated built-in function parameters. The breaking changes in question:
- Removes `user_id` input from `AddUserToUserGroup`
- Removes `channel_id` and `user_id` inputs and `user_id` output from `InviteUserToChannel`
- Removes `user_id` input from `RemoveUserFromUsergroup`
- Removes `thread_ts` input and `message_ts` output from `SendDM`
- Removes `thread_ts` input and `message_ts` output from `SendMessage`

#### Breaking Changes

A lot of the changes regarding built-ins are removing deprecated fields, so test typeahead from within `Workflow.addStep()`. Workflow steps prevent unexpected keys, so we should expect a static error if it's provided.

We should also test e2e to make sure that providing a `thread_ts` doesn't actually send that message to a thread.

**Example**
```ts
MyWorkflow.addStep(Schema.slack.functions.SendDM, {
  user_id: previousStep.output.user,
  message: "Hello darkness, my old friend",
  thread_ts: previousStep.output.message_ts,
});
```


src/schema/slack/functions/create_channel.ts
```diff
@@ -15,6 +15,12 @@ export default DefineFunction({
         description: "Enter a channel name",
         title: "Channel name",
       },
+      user_ids: {
+        type: SchemaTypes.array,
+        description: "Search all people",
```
filmaj, line 20: .. this seems like a weird field description. Just me or..?

filmaj (approved):
Looks fine to me.

How... can I test this?

src/schema/slack/functions/create_channel.ts
WilliamBergamin: That does feel off putting, maybe it should be `search all users`, not sure who to ask about changing this

shapirone: After looking again, this seems consistent with our other built-ins. The title specifies what the field does, and the description tells the user what to do. I expect this field mostly is for WFB purposes

The same description is used in
- `AddUserToUsergroup`
- `InviteUserToChannel`
- `RemoveUserFromUsergroup`

filmaj (approved):
Looks fine to me, tested and typeahead for the removed function parameters do not show up and can no longer use them in apps.

This will probably break a bunch of our samples.

src/schema/slack/functions/invite_user_to_channel_test.ts
```diff
@@ -71,11 +56,14 @@ Deno.test("InviteUserToChannel can be used as a built-in function in a workflow
     title: "Test InviteUserToChannel",
     description: "This is a generated test to test InviteUserToChannel",
   });
-  testWorkflow.addStep(InviteUserToChannel, {});
+  testWorkflow.addStep(InviteUserToChannel, {
+    channel_ids: "test",
+    user_ids: "test",
```
filmaj, line 61: Same thing here re: passing strings in but the inputs should be an array.

filmaj (approved):
LGTM, tested in apps, works fine. Left a few comments that we can probably just file as follow up issues to tackle / consider for a future date.

src/schema/slack/functions/add_user_to_usergroup_test.ts
```diff
@@ -58,11 +53,14 @@ Deno.test("AddUserToUsergroup can be used as a built-in function in a workflow s
     title: "Test AddUserToUsergroup",
     description: "This is a generated test to test AddUserToUsergroup",
   });
-  testWorkflow.addStep(AddUserToUsergroup, { usergroup_id: "test" });
+  testWorkflow.addStep(AddUserToUsergroup, {
+    usergroup_id: "test",
+    user_ids: "test",
```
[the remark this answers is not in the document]
shapirone: The answer is that technically yes it should, but we don't have static typing to enforce that in [WorkflowStepInputs](https://github.com/slackapi/deno-slack-sdk/blob/main/src/workflows/types.ts#L68). This effectively means that in practice, this test we generate is just validating pass-through and not very valuable.

I believe we have an issue around adding more typing to `WorkflowStepInputs` but it's not been prioritized. Definitely think an issue around improving the generated tests to use the right type and to use type assertions would be good!

src/schema/slack/functions/invite_user_to_channel_test.ts
WilliamBergamin: Yes this is part of the generation script behavior, If I remember correctly the types here do not actually get validated so the generation script does not implement the proper types

I originally implemented this behavior, if we want to add it in it should be possible

filmaj: I'll leave that up to @shapirone to make a call on!

PR #165, opened 2023-04-13
Remove date from built-in function files

shapirone:
###  Summary

Introduced an optional `includeDate` argument to the  `autogeneratedComment` function. This makes it so that when we update built-ins we don't get a bunch of file changes for files that weren't impacted. Left the date in the `mod.ts` file as a way for us to know the last time we ran the script (other than git history)


filmaj (approved):

WilliamBergamin (approved):
LGTM

PR #172, opened 2023-04-27
Static type testing against samples

WilliamBergamin:
###  Summary

This PR introduces a new Github workflow that executes `deno check` on a variety of [slack sample](https://github.com/slack-samples) against the changes made to the SDK in a PR.

Testing:
This PR should trigger the new workflow.

### Requirements


.github/workflows/samples.yml
```diff
@@ -0,0 +1,45 @@
...
+          path: ./deno-slack-sdk
+      - name: Checkout the ${{ matrix.sample }} sample
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ matrix.sample }}
+          path: ./sample
+
+      - name: Replace https://deno.land/x/deno_slack_sdk@x.x.x/ with ../deno-slack-sdk/src/ in import_map.json
+        working-directory: ./sample
+        run: sed -r -i 's/https:\/\/deno\.land\/x\/deno_slack_sdk@[0-9]+\.[0-9]+\.[0-9]+\//\.\.\/deno-slack-sdk\/src\//' import_map.json
```
shapirone, line 41: Will the `deno_slack_sdk@[0-9]+\.[0-9]+\.[0-9]` bit work for double digit builds like `1.10.11`?

.github/workflows/samples.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  samples:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        sample:
+          - slack-samples/deno-issue-submission
+          - slack-samples/deno-starter-template
+          - slack-samples/deno-blank-template
```
shapirone, line 18: Blank template seems like it might not be a good candidate to test as it's intentionally blank

.github/workflows/samples.yml
WilliamBergamin: Based myself off [this semver regex](https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39) but decided only to keep `major`.`minor`.`patch`

.github/workflows/samples.yml
WilliamBergamin: Makes sense, I decided to put it in because it is one of the recommended samples when executing `slack create`

shapirone: Ahh, good thought! I'm fine leaving it as it's not currently erroring. If it ends up giving us difficulties down the line we can always remove it then.

shapirone (approved):
This is awesome William! Tagged @misscoded for review of the sample coverage

.github/workflows/samples.yml
```diff
@@ -0,0 +1,45 @@
+# This workflow runs a `deno check` against slack sample apps
+name: Deno check
```
filmaj, line 2: Let's rename this, as this will be contributor-facing, so let's come up with a concise name that makes it clear what this test is doing.

```suggestion
name: Samples Integration Type-checking
```

filmaj (approved):
LGTM! Left one minor comment re: naming the test/workflow.

.github/workflows/samples.yml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/checkout@v3
+        with:
+          path: ./deno-slack-sdk
+      - name: Checkout the ${{ matrix.sample }} sample
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ matrix.sample }}
+          path: ./sample
+
+      - name: Set imports.deno-slack-sdk/ to ../deno-slack-sdk/src/ in import_map.json
```
WilliamBergamin, line 39: Opted to use a deno script over patten matching to allign with the way this Action will be implemented in the deno-slack-api

scripts/import_map/update.ts
```diff
@@ -0,0 +1,13 @@
+import { parse } from "https://deno.land/std/flags/mod.ts";
```
shapirone, line 1: We should hardcode the version for this import.

nit: Should this come from `dev_deps.ts` as well?

scripts/import_map/update.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const flags = parse(Deno.args, {
+  string: ["import-map", "sdk"],
+  default: {
+    "import-map": `${Deno.cwd()}/import_map.json`,
+    "sdk": "../deno-slack-api/src/",
```
shapirone, line 7: Since this isn't a shared script is there a reason that the default is `deno-slack-api`?

shapirone (changes requested):
Small changes with the new update script 🙇

scripts/import_map/update.ts
WilliamBergamin: > Should this come from dev_deps.ts as well?

I dont feel like it should be part of `src/dev_deps.ts` but we could create a `script/dev_deps.ts` I'm just not sure if we want to treat the `script` folder as some sort of deno package

scripts/import_map/update.ts
WilliamBergamin: Good catch 💯  that was definitely a typo originating from me trying to land the SDK and API PR simultaneously, might rethink my approach to avoid these issues

misscoded (commented):
Selected samples LGTM per previous discussion :shipit:

scripts/import_map/update.ts
WilliamBergamin: Using `script/dev_deps.ts` does not feel like the proper approach since the `script` folder is not actually a module, the scripts are individually executing for the time being

filmaj (approved):
LGTM!

shapirone (approved):
:shipit:

misscoded (approved):

PR #174, opened 2023-04-27
Add 'require_end_user_auth' property

nupurgoyall:
###  Summary
Developers of some functions may want to enforce end user auth for policy reasons. Examples include Zoom where only end user's auth should be used to start a Zoom meeting, or Salesforce where only the end user's auth is used to update a Salesforce opportunity. 

In this PR, we're adding a new property called `require_end_user_auth` for the `oauth2` input type. When this property is set, the Builder UI disable the auth delegation toggle and default to end user auth.

More context on requirements and naming in [this](https://slack-pde.slack.com/archives/C02TBKP7CAZ/p1682115306454349) thread.

[Here](https://slack-github.com/slack/webapp/pull/184865) is the webapp PR to start accepting this property.

### DIFF
--- a/src/functions/definefunction_test.ts
+++ b/src/functions/definefunction_test.ts
@@ -193,3 +193,32 @@ Deno.test("DefineFunction using an OAuth2 property requests a provider key", () 
     OAuth2Function.export().input_parameters.properties,
   );
 });
+
+Deno.test("DefineFunction using an OAuth2 property allows require_end_user_auth", () => {
+  const OAuth2Function = DefineFunction({
+    callback_id: "oauth",
+    title: "OAuth Function",
+    source_file: "functions/oauth.ts",
+    input_parameters: {
+      properties: {
+        googleAccessTokenId: {
+          type: Schema.slack.types.oauth2,
+          oauth2_provider_key: "test",
+          require_end_user_auth: true,
+        },
+      },
+      required: [],
+    },
+  });
+
+  assertEquals(
+    {
+      googleAccessTokenId: {
+        oauth2_provider_key: "test",
+        type: Schema.slack.types.oauth2,
+        require_end_user_auth: true,
+      },
+    },
+    OAuth2Function.export().input_parameters.properties,
+  );
+});
--- a/src/parameters/definition_types.ts
+++ b/src/parameters/definition_types.ts
@@ -145,6 +145,7 @@ interface NumberParameterDefinition extends BaseParameterDefinition<number> {
 interface OAuth2ParameterDefinition extends BaseParameterDefinition<string> {
   type: typeof SlackPrimitiveTypes.oauth2;
   oauth2_provider_key: string;
+  require_end_user_auth: boolean;
 }
 
 type EnumChoice<T> = {


src/functions/definefunction_test.ts
```diff
@@ -193,3 +193,32 @@ Deno.test("DefineFunction using an OAuth2 property requests a provider key", ()
     OAuth2Function.export().input_parameters.properties,
   );
 });
+
+Deno.test("DefineFunction using an OAuth2 property allows require_end_user_auth", () => {
```
filmaj, line 197: @shapirone maybe we need to tackle the TODO in the test prior to this one before we start accepting extensions to the oauth2 properties?

src/parameters/definition_types.ts
```diff
@@ -145,6 +145,7 @@ interface NumberParameterDefinition extends BaseParameterDefinition<number> {
 interface OAuth2ParameterDefinition extends BaseParameterDefinition<string> {
   type: typeof SlackPrimitiveTypes.oauth2;
   oauth2_provider_key: string;
+  require_end_user_auth: boolean;
```
filmaj, line 148: Is this a required property? That is, must users specify it?

filmaj (commented):
Left some questions and comments.

src/parameters/definition_types.ts
nupurgoyall: It's an optional property because it's not specified as required in the `oauth2` type definition in [Webapp](https://slack-github.com/slack/webapp/pull/184865/files#diff-9cf257daf9d0ce35304e8c43f85a59a5b99c72e734106d44e506766b10a976b2R3)

filmaj: Would be nice to add a JSDoc above the property as well, explaining what this option does to the developer.

nupurgoyall: Sure, I can do both! Re. making the property optional in the SDK - I wasn't able to see a difference in behavior when it was marked optional vs. not. Is there anyway to tell whether the optionality is being respected when I code up the function?

filmaj: The fact you saw no behavioural difference is probably an indication we aren't doing a very good job validating the structure of these parameter definitions on the SDK side, which is something we will work on 😬 But, even if it's not working right now, once we get to it, it helps me remember this conversation that indeed, this particular property can be optional and doesn't have to exit 😅

src/functions/definefunction_test.ts
shapirone: I don't think we should hold back this PR, but fixing that TODO would resolve the [no behavioral difference problem](https://github.com/slackapi/deno-slack-sdk/pull/174/files#r1180758629)

nupurgoyall:
@filmaj Thanks for the review. Let me know if everything looks good. If it does, I'll hold off on merging till the FE changes to consume this property are done.

nupurgoyall:
> @nupurgoyall I would say making the property optional, and explaining in a JSDoc what "end-user auth" means, is needed, then this PR is ready.

@filmaj I thought I did both of those things (based on the commit message) as part of [this commit](https://github.com/slackapi/deno-slack-sdk/pull/174/commits/b6f59843db39df0868c1f3682f692ac3bf2e31ee) but clearly forgot to make the property optional 🤦‍♀️. Does the docstring in that commit answer your questions? If not, I'm happy to go into more details but I also wanted to be as succinct as possible and not call out specific providers in the docstring.

> Another way to explore this question: would our [deno-github-functions sample](https://github.com/slack-samples/deno-github-functions#register-a-github-app) need to set this property to true or false?

It's up to the developer to set this for Github. If they don't set it, then the workflow builder can choose whether to use the builder's or end user's credentials. We're making this property available for cases where not using end user's auth would violate some policy. For example, Pagerduty has told us that they would want all workflows to be run with the end user's auth for monetization purposes.

src/parameters/definition_types.ts
```diff
@@ -147,7 +147,7 @@ interface OAuth2ParameterDefinition extends BaseParameterDefinition<string> {
   /** Specifies the oauth2 provider this input is associated to */
   oauth2_provider_key: string;
   /** Dictates whether only the auth of the user running the workflow should be used */
```
filmaj, line 149: This defaults to `false`, yes? If so, let's clarify that:
```suggestion
  /** Dictates whether only the auth of the user running the workflow should be used. Defaults to `false`. */
```

filmaj (approved):

filmaj:
Once I get testing instructions, I will give this a spin and possibly offer suggestions/tweaks to the JSDoc.
Let's also hold off on merging until @nupurgoyall confirms it is ready for consumption in production.

nupurgoyall:
@shapirone @filmaj Are you okay with me merging this PR now. The FE changes are now in place; however, the "Samples Integration Type-checking" test is failing with `error: TS2352 [ERROR]: Conversion of type 'ValidTriggerResponseObjects[]' to type 'ReactionTriggerResponseObject[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.` which seems unrelated.

filmaj:
@nupurgoyall no worries about the test failure, [that particular sample is waiting on a dependency upgrade](https://github.com/slack-samples/deno-simple-survey/pull/20) (deno-slack-api 2.1.0 -> 2.1.1) to be merged. This repo, however, has deno-slack-api 2.1.1. already integrated. So this failure is just calling out this mismatch/incompatibility between using different versions of the deno-slack-api.

I verified locally that bumping the deno-slack-api version to 2.1.1 in that sample, and then integrating with this PR, passes the typechecks just fine.

Gonna merge this shortly.

PR #176, opened 2023-05-16
Add ConnectorDefinition

WilliamBergamin:
###  Summary

This PR introduces a `ConnectorDefinition` object, it is very similar to the `SlackFunctionDefinition` object
Some refactoring was put in place in order to reuse code and improve (I think) the type system.

Looking for feedback on:
- The use of inheritance for `SlackFunction`, `Connectors` and eventually `builtins`, this approach was chosen to maintain backwards compatibility and create no disturbance  with `SlackFunctionDefinition`
- The use of an abstract class `BaseFunctionDefinition` that implements `ISlackFunctionDefinition` to define `SlackFunctionDefinition` and `ConnectorDefinition`.
- Should other parts of the SDK reference `BaseFunctionDefinition` instead of `SlackFunctionDefinition`?
  - From what I've experimented with only `SlackFunctionDefinition` should be referenced, `ISlackFunctionDefinition` is what is used in workflow steps.
- A new top level export file `src/internal.ts` was added, this file aims to export components that are used by other slack packages and not intended for public consumption.
- Naming
  - `BaseFunctionDefinition` could be `FunctionDefinition`?
  - `ConnectorDefinition` could be `SlackConnectorDefinition` but they are not actually slacks connectors
  - `SlackFunctionDefinitionArgs` and `FunctionDefinitionArgs`

### Requirements


src/functions/mod.ts
```diff
@@ -1,9 +1,16 @@
...
+  ManifestFunctionSchema,
+  ManifestFunctionType,
+} from "../manifest/manifest_schema.ts";
 import {
   ParameterSetDefinition,
   PossibleParameterKeys,
 } from "../parameters/types.ts";
-import { FunctionDefinitionArgs } from "./types.ts";
+import {
+  FunctionDefinitionArgs,
```
shapirone, line 10: nit: Should we rename to `BaseFunctionDefinitionArgs`?

shapirone (commented):
Looking good! Just did a really quick review with some questions/suggestions. As far as your questions, the only one I don't think I answered in-line is

> Should other parts of the SDK reference BaseFunctionDefinition instead of SlackFunctionDefinition?

I don't think so, the following are what leverage `SlackFunctionDefinition` today and only have use for them:
- `DefineFunction`
- `SlackFunction`
- `SlackFunctionTester`
- Action/View Handlers

That being said, we may want to consider renaming `ISlackFunctionDefinition` since it's being used for more than just SlackFunctions

src/manifest/manifest_schema.ts
```diff
@@ -207,10 +207,13 @@ export type ManifestFunction = ISlackFunctionDefinition<any, any, any, any>;
 
 export type ManifestFunctionsSchema = { [key: string]: ManifestFunctionSchema };
 
+export type ManifestFunctionType = "API" | "builtin" | "app" | undefined;
```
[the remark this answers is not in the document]
WilliamBergamin: 💯  this makes sense if we want to add `builtin` in the future we will be able

src/internal.ts
```diff
@@ -0,0 +1,2 @@
+export { InternalSlackTypes } from "./schema/slack/types/custom/mod.ts";
+export { ConnectorDefinition } from "./functions/mod.ts";
```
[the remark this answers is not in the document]
WilliamBergamin: Sounds good 👍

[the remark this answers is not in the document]
WilliamBergamin: I originally created this because `ConnectorDefinition` is not referenced anywhere, it looks like its not used at all from this package perspective

src/functions/mod.ts
```diff
@@ -66,3 +106,39 @@ export class SlackFunctionDefinition<
     };
   }
 }
+
+export class ConnectorDefinition<
```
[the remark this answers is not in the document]
WilliamBergamin: Adding a `DefineConnector` function felt redundant, we can just do `new ConnectorFunctionDefinition({})` instead
Lmk what you think

src/functions/mod.ts
```diff
@@ -66,3 +106,39 @@ export class SlackFunctionDefinition<
...
+      InputParameters,
+      OutputParameters,
+      RequiredInput,
+      RequiredOutput
+    >,
+  ) {
+    super(definition);
+  }
+
+  export(): ManifestFunctionSchema {
```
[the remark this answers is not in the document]
WilliamBergamin: They are slightly different, the `ConnectorFunctionDefinition` `export` has an additional `type` field, adding this to the base class would change the output of `SlackFunctionDefinition.export()`

src/functions/mod.ts
WilliamBergamin: I'm not a huge fan of the `base` prefix, if we don't need it I would try not to use it

src/functions/mod.ts
```diff
@@ -17,7 +24,7 @@ export const DefineFunction = <
   RequiredInput extends PossibleParameterKeys<InputParameters>,
   RequiredOutput extends PossibleParameterKeys<OutputParameters>,
 >(
-  definition: FunctionDefinitionArgs<
+  definition: SlackFunctionDefinitionArgs<
```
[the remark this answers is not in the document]
WilliamBergamin: 💯 I changed the file structure of the module to isolate the `FunctionDefinition` objects
Lmk what you think

Might change the unit tests in a follow up PR

WilliamBergamin:
Any optionions on renaming the abstract class `BaseFunctionDefinition` -> `FunctionDefinition`?

src/functions/mod.ts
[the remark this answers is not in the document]
shapirone: Our `Define*` methods instantiating and returning the class instance is our established pattern. Let's stick to it.

This pattern pre-dates me and I never discussed it in depth, but I always assumed it was because functions are a more common pattern in JS libs than classes

src/functions/mod.ts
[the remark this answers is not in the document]
shapirone: Aren't we assigning `type: "app"` in `SlackFunctionDefinition`, meaning that we can return the `type` on those functions in the output manifest?

src/functions/mod.ts
shapirone: Yeah the `base` prefix is an established pattern we use pretty commonly in this repo and others to indicate a parent class. In this case my thinking was that because these are the args for `BaseFunctionDefinition`, not necessarily the args for `SlackFunctionDefinition` or `ConnectorFunctionDefinition`, and it seems nice to indicate as such in the name

src/functions/mod.ts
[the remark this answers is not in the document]
shapirone: Hmm, the new file moves `SlackFunctionDefinition` but doesn't move `DefineFunction` which is what instantiates the class.

I might consider
- `base_function.ts` -> export `BaseFunctionDefinition`
- `slack_function.ts` -> export `SlackFunctionDefinition` and `DefineFunction`
- `connector_function.ts` -> export `ConnectorFunctionDefinition` and `DefineConnector`
- `mod.ts` -> re-export `DefineFunction` and `DefineConnector`

src/internal.ts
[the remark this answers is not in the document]
shapirone: Yeah, but that's temporary — eventually we'll export `DefineConnector` from our `src/mod.ts` file

src/functions/mod.ts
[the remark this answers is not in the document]
WilliamBergamin: Currently `type: "app"` is in `SlackFunctionDefinition` but is not part of the `returnType` of `export` adding `type: "app"` to the `returnType` of `export` may affect some users and mays also affect some tests related to the `generated builtin`

We did talk about remove the affected tests in the past 🧠 

I would recommend adding `type: "app"` to the `returnType` of `export` and removing the affected tests in a separate PR?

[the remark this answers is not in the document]
shapirone: But in this PR aren't you adding support for "app" [in the `returnType` of `export`](https://github.com/slackapi/deno-slack-sdk/pull/176/files#diff-8d25a59e46a06143e923f890c73ca8de7fa24f95ed7bc8bac40219bc78ba17afR210)? We just never actually export it from Slack functions, only from Connector functions. I'm not sure what the purpose of assigning `type: "app"` in `SlackFunctionDefinition` is if we don't use it 🤔 I assumed that returning it in the output manifest would work because it is the default if no type is specified?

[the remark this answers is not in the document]
WilliamBergamin: Yes it does work, my plan was to add `type: "app"` to the export in a follow up PR, I tried to keep this PR as simple as possible and not change the SDK behavior at all

I would also want to create a `BuiltinFunctionDefinition` in this follow up PR and use it in the generation script 🏄 
Lmk what you think

src/functions/mod.ts
WilliamBergamin: Makes sense, I've renamed the type 🟢

WilliamBergamin: Renaming `FunctionDefinitionArgs` to `BaseFunctionDefinitionArgs` introduced many changes to in interactivity folder which does not seem necessary
I would potentially want to rename the following 

- `BaseFunctionDefinition` -> `FunctionDefinition`
- `BaseFunctionDefinitionArgs` -> `FunctionDefinitionArgs`

I feel like this is more descriptive and introduces less change in the PR, I will implement it now to keep the PR clean let me know what you think

src/functions/mod.ts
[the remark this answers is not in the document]
shapirone: Curious what we'll get from `BuiltinFunctionDefinition` — if there's value it may make sense to keep that in the repo where built-ins exist long-term and not necessarily the SDK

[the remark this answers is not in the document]
WilliamBergamin: I see, this is a classic java pattern where the SDK defines an interface and the other packages implements it

tests/integration/functions/define_function_test.ts
```diff
@@ -1,6 +1,5 @@
-import { DefineFunction } from "./mod.ts";
-import Schema from "../schema/mod.ts";
-import { assertEquals, assertStrictEquals } from "../dev_deps.ts";
+import { assertEquals, assertStrictEquals } from "../../../src/dev_deps.ts";
```
shapirone, line 1: Why was this moved to be an integration test? AFAIK this file is performing unit tests on `DefineFunction` which is really testing the underlying `SlackFunctionDefinition` class logic.

@filmaj can correct me but my understanding is our integration files are files that are intentionally testing two separate imports together

My two cents is to just test `DefineFunction` and `DefineConnector`, which will execute the corresponding function definition

src/manifest/manifest_schema.ts
```diff
@@ -207,10 +207,13 @@ export type ManifestFunction = ISlackFunctionDefinition<any, any, any, any>;
...
 export type ManifestFunctionsSchema = { [key: string]: ManifestFunctionSchema };
 
+export type ManifestFunctionType = "API" | "app" | undefined;
+
 export type ManifestFunctionSchema = {
+  type?: ManifestFunctionType;
   title?: string;
   description?: string;
-  source_file: string;
+  source_file?: string;
```
shapirone, line 216: `ManifestFunctionSchema` is the shape of the manifest's functions object. Since we are focused on the consumption of Connectors at the moment, we don't need to support them being added to the manifest yet. I'm inclined to not change this yet so that devs aren't able to register Connectors to their own manifests before we fully support it.

src/functions/definitions/base_function.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.id = definition.callback_id;
+    this.definition = definition;
+  }
+
+  registerParameterTypes(manifest: SlackManifest) {
+    const { input_parameters: inputParams, output_parameters: outputParams } =
+      this.definition;
+    manifest.registerTypes(inputParams?.properties ?? {});
+    manifest.registerTypes(outputParams?.properties ?? {});
+  }
```
shapirone, line 44: `registerParameterTypes` is intended to find any types used by a function and register them to the manifest if they're not already there. This seems like something we'd only need to do on `SlackFunction`

src/functions/definitions/base_function_test.ts
```diff
@@ -0,0 +1,97 @@
...
+    "out",
+  ]);
+  assertEquals(
+    AllTypesFunction.definition.input_parameters?.properties.myString.hint,
+    "Ex. my neat value",
+  );
+  assertEquals(
+    AllTypesFunction.definition.input_parameters?.properties.myBoolean.hint,
+    "Ex: true/false",
+  );
```
shapirone, line 96: nit: Add assertion to validate `id` is what we expect?

src/functions/definitions/base_function_test.ts
```diff
@@ -0,0 +1,97 @@
...
+  ManifestFunctionSchema,
+  ManifestFunctionType,
+} from "../../manifest/manifest_schema.ts";
+import {
+  ParameterSetDefinition,
+  PossibleParameterKeys,
+} from "../../parameters/types.ts";
+import { FunctionDefinitionArgs } from "../types.ts";
+
+export class TestBaseFunctionDefinition<
```
shapirone, line 15: nit: I wonder how valuable it is to make `FunctionDefinition` an abstract class when we need to do something like this to test.

Theoretically the majority of our tests should be testing the base `FunctionDefinition` functionality, and tests on the extended classes should focus on the extensions since we validated what `FunctionDefinition` does

src/functions/definitions/connector_function_test.ts
```diff
@@ -0,0 +1,80 @@
+import { assertEquals, assertStrictEquals } from "../../dev_deps.ts";
```
shapirone, line 1: A lot of the tests in this file validate `FunctionDefinition` logic, rather than `ConnectorFunctionDefinition`

src/functions/definitions/mod_test.ts
```diff
@@ -0,0 +1,42 @@
+import { SlackFunctionDefinition } from "./slack_function.ts";
+import { ConnectorFunctionDefinition } from "./connector_function.ts";
+import { DefineConnector, DefineFunction } from "./mod.ts";
+import { assertInstanceOf } from "../../dev_deps.ts";
+import { PossibleParameterKeys } from "../../parameters/types.ts";
+
+type emptyParameterType = Record<string, never>;
+
+Deno.test("DefineFunction sets appropriate defaults", () => {
```
shapirone, line 9:
```suggestion
Deno.test("DefineFunction returns an instance of `SlackFunctionDefinition`", () => {
```

src/functions/definitions/mod_test.ts
```diff
@@ -0,0 +1,42 @@
...
+    SlackFunctionDefinition<
+      emptyParameterType,
+      emptyParameterType,
+      PossibleParameterKeys<emptyParameterType>,
+      PossibleParameterKeys<emptyParameterType>
+    >,
+  );
+});
+
+Deno.test("DefineConnector sets appropriate defaults", () => {
```
shapirone, line 27:
```suggestion
Deno.test("DefineConnector returns an instance of ConnectorFunctionDefinition", () => {
```

src/functions/definitions/slack_function_test.ts
```diff
@@ -0,0 +1,85 @@
+import { SlackFunctionDefinition } from "./slack_function.ts";
```
shapirone, line 1: A lot of the tests in this file validate `FunctionDefinition` logic, rather than `SlackFunctionDefinition`

shapirone (changes requested):
Excellent! My comments are most around tests and the only change I think we need to make is to remove the changes to the manifest's functions object so that devs don't have the ability to add a connector to their manifest before we actually support it.

There's also a bit of casing inconsistency for file names, but honestly we've been fairly inconsistent. I'll bring this up in Slack for the group to decide what we want to do

src/manifest/manifest_schema.ts
WilliamBergamin: Should we set `source_file=""` for connectors or create a new `type`?

shapirone: Do we need to if connectors don't have an export?

tests/integration/functions/define_function_test.ts
WilliamBergamin: I decided to move `define_function_test.ts` here since it seem contain more then just "pure unite tests", this file seems to validate the full behavior of `DefineFunction`

Example: 

- "DefineFunction with required params"
- "DefineFunction using an OAuth2 property allows require_end_user_auth"

Since theses tests cover all the new changes (`FunctionDefinition` abstract class + SlackFunctionDefinition) I thought they resembled more integration tests rather then "pure unite" tests

shapirone: Both of those examples stem from the underlying implementation thoug, rather than coming from `DefineFunction` itself. I'd think we should just move them to the corresponding file?

src/functions/definitions/base_function.ts
WilliamBergamin: Ahh I see this makes a lot of sense 💯  let me try out removing this from the base class

src/functions/definitions/connector_function.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      title: this.definition.title,
+      description: this.definition.description,
+      type: this.type,
+      input_parameters: this.definition.input_parameters ??
+        { properties: {}, required: [] },
+      output_parameters: this.definition.output_parameters ??
+        { properties: {}, required: [] },
+    };
+  }
```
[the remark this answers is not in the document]
WilliamBergamin: Ah 🤔  the reason why `export()` is in the abstract base class and `ConnectorFunctions` implements it is because `addStep` expects [ISlackFunctionDefinition](https://github.com/slackapi/deno-slack-sdk/blob/156def3d584b6b66a8e3935761b508188d8494fe/src/functions/types.ts#LL302C30-L302C30) as an [argument](https://github.com/slackapi/deno-slack-sdk/blob/156def3d584b6b66a8e3935761b508188d8494fe/src/workflows/mod.ts#L119)

We can potentially change this but it will be a bigger refactor then I expected, LMK what you think

src/functions/definitions/base_function.ts
WilliamBergamin: I figured out why I implemented `registerParameterTypes` at the a abstract class level 💡the reason for this is that  `addStep` expects [ISlackFunctionDefinition](https://github.com/slackapi/deno-slack-sdk/blob/156def3d584b6b66a8e3935761b508188d8494fe/src/functions/types.ts#LL302C30-L302C30) as an [argument](https://github.com/slackapi/deno-slack-sdk/blob/156def3d584b6b66a8e3935761b508188d8494fe/src/workflows/mod.ts#L119)

We can potentially change this but it will be a bigger refactor then anticipated, LMK what you think

WilliamBergamin: The `deno-slack-runtime` filters out the functions of `type: API` from the manifest

src/manifest/manifest_schema.ts
WilliamBergamin: Seems like we may need to implement `export` in `ConnectorFunctionDefinition` I expand on this in [this comment](https://github.com/slackapi/deno-slack-sdk/pull/176#discussion_r1214698723)

src/functions/definitions/base_function_test.ts
WilliamBergamin: done 🟢

src/functions/definitions/connector_function.ts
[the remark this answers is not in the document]
shapirone: Hmm, are there negative repercussions making `export` in `ISlackFunctionDefinition` optional?

src/manifest/manifest_schema.ts
WilliamBergamin: We would have to add the following where ever `export` or `registerParameterTypes` is used

```ts
if (fn.export) {
}

if (fn.registerParameterTypes) {
}
```

They seem to only be used in 2 places

src/functions/definitions/connector_function.ts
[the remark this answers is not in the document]
WilliamBergamin: We would have to add the following where ever `export` or `registerParameterTypes` is used

```ts
if (fn.export) {
}

if (fn.registerParameterTypes) {
}
```

They seem to only be used in 2 places, maybe this is worth implementing? lmk what you think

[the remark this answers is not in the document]
shapirone: That seems acceptable to me!

tests/integration/functions/define_function_test.ts
WilliamBergamin: I've moved these test back to the unit test file

src/functions/definitions/base_function.ts
WilliamBergamin: Edit this was in the [deno-slack-hooks build.ts](https://github.com/slackapi/deno-slack-hooks/blob/main/src/build.ts#L32)

But I've modified the SDK so `connectors` are not exported in the manifest

src/functions/definitions/base_function_test.ts
WilliamBergamin: I've removed the use of a `base` abstract class
- The logic was moved to SlackFunctionDefinition
- ConnectorDefinitions are not included in the Manifest and don't require this logic

src/functions/definitions/connector_function_test.ts
WilliamBergamin: The abstract base class was no longer required I've removed it and this test file 💯

src/functions/definitions/slack_function_test.ts
WilliamBergamin: The abstract base class was no longer required I've removed it and its test file 💯

src/functions/definitions/slack-function_test.ts
```diff
@@ -222,3 +251,34 @@ Deno.test("DefineFunction using an OAuth2 property allows require_end_user_auth"
...
+    callback_id: "no_params",
+    title: "No Parameter Function",
+    source_file: "functions/no_params.ts",
+  });
+
+  assertInstanceOf(NoParamFunction, SlackFunctionDefinition);
+  assertEquals(true, isCustomFunctionDefinition(NoParamFunction));
+});
+
+Deno.test("isCustomFunction should return false when a non custom function is passed", () => {
```
shapirone, line 266: nit:
```suggestion
Deno.test("isCustomFunctionDefinition should return false when a non custom function is passed", () => {
```

src/functions/definitions/slack-function_test.ts
```diff
@@ -222,3 +251,34 @@ Deno.test("DefineFunction using an OAuth2 property allows require_end_user_auth"
     OAuth2Function.export().input_parameters.properties,
   );
 });
+
+Deno.test("isCustomFunction should return true when SlackFunctionDefinition is passed", () => {
```
shapirone, line 255: nit:
```suggestion
Deno.test("isCustomFunctionDefinition should return true when SlackFunctionDefinition is passed", () => {
```

shapirone (approved):
🚀 LGTM! 🚀

PR #178, opened 2023-05-26
Bump the deno-slack-api version

WilliamBergamin:
###  Summary

This PR simply bumps the deno-slack-api version to the latest

### Requirements

### DIFF
--- a/src/deps.ts
+++ b/src/deps.ts
@@ -1,5 +1,5 @@
-export { SlackAPI } from "https://deno.land/x/deno_slack_api@2.1.0/mod.ts";
+export { SlackAPI } from "https://deno.land/x/deno_slack_api@2.1.1/mod.ts";
 export type {
   SlackAPIClient,
   Trigger,
-} from "https://deno.land/x/deno_slack_api@2.1.0/types.ts";
+} from "https://deno.land/x/deno_slack_api@2.1.1/types.ts";


filmaj (approved):

filmaj (changes requested):
Wait! Some samples integration tests failed!

filmaj:
@WilliamBergamin when I updated the sample's import of deno-slack-api to the latest version, then this error goes away.

The fact that developers have to import both SDK and API, and the SDK includes an API version in it as well, is troublesome. Easy to get out of sync.

WilliamBergamin:
Ohh I see, when I tried to update deno-slack-api it was still breaking but I was using the old version of the SDK.

Maybe would should not tell devs to import the API independently maybe the SDK should export it and users should use that one by default?

WilliamBergamin:
The [action found](https://github.com/slackapi/deno-slack-api/actions/runs/4887786928) related to the PR that caused this github action to fail are passing.

The reason behind this is that the Github action found in the deno-slack-api repo properly substitutes the deno-slack-api version in the sample project  as well as in the deno-slack-sdk package. The action present in this package only substitutes the deno-slack-sdk version in the sample project without updating the deno-slack-api version

For this reason I will merge this PR and we will address the issue of package desynchronization in the near future

PR #189, opened 2023-07-10
[HERMES-5354] Add `userScopes` to `ISlackManifestShared`

srajiang:
###  Summary
Adds the `userScopes` property to the ISlackManifestShared interface.
Ask from Hermes User Delegated Auth folks to enable authoring app manifest that use built-ins with user-delegated auth. 

[Context thread](https://slack-pde.slack.com/archives/C0478C43T4J/p1687192300541109)

Changes: 
* Moves `userScopes` property to the `ISlackManifestShared` interface
* Sets the value
* Normally Run on Slack apps have `token_management_enabled` defaulted to false on behalf of the developer. But in the case where user scopes are requested, the backend requires `token_management_enabled` to be set to `true`. So I have modified the SDK behavior to default to setting the value to true only when user scopes are defined. 

### DIFF
--- a/src/manifest/manifest_test.ts
+++ b/src/manifest/manifest_test.ts
@@ -745,6 +745,32 @@ Deno.test("Manifest() correctly assigns other app features", () => {
   );
 });
 
+Deno.test("Manifest() ensures that apps defining user-level scopes also manage tokens", () => {
+  const definition: SlackManifestType = {
+    runOnSlack: false,
+    name: "fear and loathing in las vegas",
+    description:
+      "fear and loathing in las vegas: a savage journey to the heart of the american dream",
+    displayName: "fear and loathing",
+    icon: "icon.png",
+    botScopes: ["channels:history", "chat:write", "commands"],
+    userScopes: ["users:read"],
+  };
+  const manifest = Manifest(definition);
+
+  // ensure user scopes are set on the manifest
+  assertStrictEquals(
+    manifest.oauth_config.scopes.user,
+    definition.userScopes,
+  );
+
+  // ensure that token management is set to true
+  assertStrictEquals(
+    manifest.oauth_config.token_management_enabled,
+    true,
+  );
+});
+
 Deno.test("SlackManifest() registration functions don't allow duplicates", () => {
   const functionId = "test_function";
   const arrayTypeId = "test_array_type";
--- a/src/manifest/mod.ts
+++ b/src/manifest/mod.ts
@@ -50,6 +50,7 @@ export class SlackManifest {
       oauth_config: {
         scopes: {
           bot: this.ensureBotScopes(),
+          user: def.userScopes,
         },
       },
       features: {
@@ -256,10 +257,9 @@ export class SlackManifest {
     manifest.app_directory = def.appDirectory;
 
     //OauthConfig
-    manifest.oauth_config.scopes.user = def.userScopes;
     manifest.oauth_config.redirect_urls = def.redirectUrls;
 
-    // Remote-hosted Slack apps manage their own tokens
+    // Remote-hosted Slack apps always manage their own tokens
     manifest.oauth_config.token_management_enabled = true;
 
     // Remote Features
@@ -274,9 +274,11 @@ export class SlackManifest {
   private assignRunOnSlackManifestProperties(manifest: ManifestSchema) {
     const def = this.definition as ISlackManifestRunOnSlack;
 
-    // Run on Slack Apps do not manage access tokens
-    // This is set by default as false
-    manifest.oauth_config.token_management_enabled = false;
+    // Oauth Config
+    // Run On Slack manage their own tokens
+    // if and only if user scopes are included in the manifest
+    manifest.oauth_config.token_management_enabled =
+      manifest.oauth_config.scopes.user != undefined;
 
     // Required App Settings for run on slack apps
     manifest.settings.org_deploy_enabled = true;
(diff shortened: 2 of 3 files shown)


src/manifest/types.ts
```diff
@@ -60,26 +60,26 @@ export interface ISlackManifestRemote extends ISlackManifestShared {
   socketModeEnabled?: boolean;
   tokenRotationEnabled?: boolean;
   appDirectory?: ManifestAppDirectorySchema;
-  userScopes?: Array<string>;
```
srajiang, line 63: No longer needed here since we're moving it up to the shared interface.

src/manifest/mod.ts
```diff
@@ -274,9 +274,11 @@ export class SlackManifest {
...
     const def = this.definition as ISlackManifestRunOnSlack;
 
-    // Run on Slack Apps do not manage access tokens
-    // This is set by default as false
-    manifest.oauth_config.token_management_enabled = false;
+    // Oauth Config
+    // Run On Slack manage their own tokens
+    // if and only if user scopes are included in the manifest
+    manifest.oauth_config.token_management_enabled =
+      manifest.oauth_config.scopes.user != undefined;
```
filmaj, line 281: If the user specifies user scopes as an empty array, token management will be enabled. Is this the behaviour we want?

src/manifest/mod.ts
```diff
@@ -50,6 +50,7 @@ export class SlackManifest {
       oauth_config: {
         scopes: {
           bot: this.ensureBotScopes(),
+          user: def.userScopes,
```
filmaj, line 53: How does the API respond to user scopes being set to an empty array?

filmaj (commented):
Left a couple of questions.

src/manifest/manifest_test.ts
```diff
@@ -745,6 +745,32 @@ Deno.test("Manifest() correctly assigns other app features", () => {
...
+    definition.userScopes,
+  );
+
+  // ensure that token management is set to true
+  assertStrictEquals(
+    manifest.oauth_config.token_management_enabled,
+    true,
+  );
+});
+
```
WilliamBergamin, line 773: Seems like this file defines ` botScopes: []` in many tests 🤔 
Should we also define `userScopes:[]` in those tests as well @filmaj ?

WilliamBergamin (commented):
Good changes 💯 just had one query

src/manifest/manifest_test.ts
filmaj: I'd like for the tests to use `definition` variables that reflect how end-users would craft their own manifest definitions. In my own experience creating manifest definitions, I would omit the property altogether if I'm not using it. However, devs do the darndest things! I think we should guard against common JS-y/TS-y pitfalls such as what counts as truthy vs. falsy, like empty arrays and empty strings. Therefore, I would recommend the `Manifest()` constructor clean user-provided definitions up into a state that the Slack manifest APIs expect as a baseline.

src/manifest/mod.ts
```diff
@@ -274,9 +274,17 @@ export class SlackManifest {
   private assignRunOnSlackManifestProperties(manifest: ManifestSchema) {
     const def = this.definition as ISlackManifestRunOnSlack;
 
-    // Run on Slack Apps do not manage access tokens
-    // This is set by default as false
-    manifest.oauth_config.token_management_enabled = false;
+    // Oauth Config
+    // Run On Slack manage their own tokens
+    // if and only if at least one user scope is included in the manifest
```
jpbrabec, line 279: I'm not sure I understand this-- Adding scopes in the manifest should not require apps to use token management. In general I don't think any ROSI apps needs token management, right? Having token management enabled would mean the app needs to provide an oauth token server, which ROSI apps cannot do.

src/manifest/mod.ts
srajiang: Tried to force a token_management_enabled=true + oauthconfig.scopes.user = [] scenario to see what the backend says, but it seems like today due to the way the go marshals empty arrays (omitempty), it gets removed entirely from the payload sent to Slack. Kind of surprising behavior, since i guess we could technically pass this kind of test at the SDK level, but it wouldn't be reflective of what the CLI sends over to Slack.

```

  const definition2: SlackManifestType = {
    runOnSlack: true,
    name: "fear and loathing in las vegas",
    description:
      "fear and loathing in las vegas: a savage journey to the heart of the american dream",
    displayName: "fear and loathing",
    icon: "icon.png",
    botScopes: ["channels:history", "chat:write", "commands"],
    userScopes: [], // empty user scopes
  };
  const manifest2 = Manifest(definition2);

  assertStrictEquals(
    manifest2.oauth_config.scopes.user,
    definition2.userScopes,
  );
```

src/manifest/mod.ts
srajiang: @filmaj - Good catch, thank you! token management enabled should remain `false` in that case for run on slack apps. I've updated the SlackManifest + the tests.

src/manifest/mod.ts
srajiang: @jpbrabec - This was also something I'm also unsure about, so thanks for weighing in.

> Adding scopes in the manifest should not require apps to use token management.

I get this error when adding user scopes to the manifest and leaving token_management_enabled: false: 
Screenshot 2023-07-09 at 10 04 52 PM

Is this error expected?

jpbrabec: Ah, that's for user scopes specifically (not bot scopes). Hermes apps do not support user scopes *at all* right now, so this error is basically here to say "You can't be a hermes ROSI app and use user scopes at the same time".

slackmon: @jpbrabec are you open to me changing the error to be able to bypass it if Hermes User Delegated Auth Houston FF is in place

PR #192, opened 2023-07-26
Error on duplicate assets

WilliamBergamin:
###  Summary

This PR introduces new behavior that loudly errors when a user tries to create a manifest with duplicate assets
- workflow
- function
- custom type
- datastores
- custom events
- OAuth providers

This error will pop up whenever a user executes a command that uses the `slack manifest` hook

#### Feedback

I'm looking for feedback on
- the approach, is there a cleaner way to implement this?
- naming
- behavior, is this the behavior we want for this issue

#### Testing

1. create a project
2. place the following in the import map `"deno-slack-sdk/": "https://raw.githubusercontent.com/slackapi/deno-slack-sdk/e4c2c2b1eb0b8f604b74be5c36f224bd1b56068b/src/"`
3. create a 2 workflows with the same `callback_id`
4. run `slack manifest`

#### Sample output

```zsh
error: Uncaught (in promise) Error: Duplicate callback_id: "submit_issue" found in Workflow.
throw new DuplicateCallbackIdError(workflow.id, "Workflow");
                  ^
at https://raw.githubusercontent.com/slackapi/deno-slack-sdk/e4c2c2b1eb0b8f604b74be5c36f224bd1b56068b/src/manifest/mod.ts:87:19
at Array.reduce (<anonymous>)
at SlackManifest.export (https://raw.githubusercontent.com/slackapi/deno-slack-sdk/e4c2c2b1eb0b8f604b74be5c36f224bd1b56068b/src/manifest/mod.ts:84:42)
at Manifest (https://raw.githubusercontent.com/slackapi/deno-slack-sdk/e4c2c2b1eb0b8f604b74be5c36f224bd1b56068b/src/manifest/mod.ts:32:19)
at file:///Users/wbergamin/Documents/slack/tools/experiments/frosty-bear-975/manifest.ts:16:16
Check /Users/wbergamin/.slack/logs/slack-debug-20230802.log for full error logs

🚫 Unable to get your app manifest details. Please check your manifest file. (invalid_manifest)

Error Details:

1: Command for 'GetManifest' returned an error: exit status 1 (sdk_hook_invocation_failed)
Suggestion: Run `slack doctor` to check that your system dependencies are up-to-date.
```

### Requirements


src/manifest/manifest_schema.ts
```diff
@@ -296,10 +297,7 @@ export type ManifestOAuth2Schema = {
 
 export type ManifestOAuth2ProviderSchema = {
   provider_type: OAuth2ProviderTypeValues;
-  options: {
-    // deno-lint-ignore no-explicit-any
-    [key: string]: any;
-  };
+  options: OAuth2ProviderOptions;
```
WilliamBergamin, line 300: This change may be breaking 😅 
The `DefineOAuth2Provider` function requires its `options` parameter to be of type `OAuth2ProviderOptions`, but this may break apps that are defining there `ManifestOAuth2ProviderSchema` manually?

WilliamBergamin (commented):
Maybe a breaking change here, but I could remove it from this PR

src/manifest/errors.ts
```diff
@@ -0,0 +1,160 @@
+import {
+  ManifestCustomEventSchema,
+  ManifestCustomTypeSchema,
+  ManifestDatastoreSchema,
+  ManifestFunctionSchema,
+  ManifestOAuth2ProviderSchema,
+  ManifestWorkflowSchema,
+} from "./manifest_schema.ts";
+
+export class DuplicateWorkflowError extends Error {
```
filmaj, line 10: What do you think about reducing the number of errors here down to a set of `Duplicate<PropertyName>Error`? I.e. `DuplicateCallbackIdError` and `DuplicateNameError`. Then, each of these can be passed the various Slack constructs that require unique strings for their relevant properties. That is:

- `DuplicateCallbackIdError` can apply to either workflows or functions, and the output from this error makes clear when the error applies to functions and when to workflows.
- `DuplicateNameError` can apply to datastores, events, types, etc, and again the output from this error makes clear when the error applies to which construct.

src/manifest/errors.ts
```diff
@@ -0,0 +1,160 @@
...
+    id: string,
+    current: ManifestWorkflowSchema,
+    old: ManifestWorkflowSchema,
+  ) {
+    super([
+      `Duplicate callback_id: "${id}" for Workflow`,
+      DuplicateWorkflowError.stringifySimpleSchema(current),
+      "---",
+      DuplicateWorkflowError.stringifySimpleSchema(old),
+    ].join("\n"));
```
filmaj, line 31: The output is a little weird:

```
error: Uncaught (in promise) Error: Duplicate callback_id: "create_new_issue_workflow" for Workflow
{
"title": "Create new issue2",
"description": "Create a new GitHub issue"
}
---
{
"title": "Create new issue",
"description": "Create a new GitHub issue"
}
```

Do we have to show the error data as JSON? And I think displaying title and description is not ideal, as I often duplicate these fields (and they don't need to be unique, either). Can we instead display the path to the relevant source files containing the duplicate IDs?

filmaj (changes requested):
Hey awesome work! I totally forgot that this same class of problems exist for all of our various types, like events, custom types, etc. Nicely done.

I left a few questions and suggestions. It's also unclear to me what you are thinking in terms of making changes in deno-slack-hooks related to this issue?

src/manifest/errors.ts
WilliamBergamin: I agree, I was on the fence about what better way to implement this, looking back this approach should reduce the amount of boiler plate cope

src/manifest/errors.ts
WilliamBergamin: We do not need to show the error data as JSON, I opted for this solution since I felt like like it provided the data in a clean way, maybe a table pattern is better

> Can we instead display the path to the relevant source files containing the duplicate IDs?

Figuring out what data to display to the user in order to trace down what workflows are duplicated has proven to be a tricky one, for functions it is easy to display the `source_file` field but for other assets like workflows or customTypes they don't have this field and could be defined anywhere in the project

I opted to display the `title` and `description` because it provides some information to the user, my logic was that if a user copy and pastes a workflow and forget to change the `callback_id`, `title` and `description`, displaying the  `title` and `description` might remind them of which workflow they copy and pasted

@filmaj what did you have in mind to fetch the source_files of the objects?

filmaj: > maybe a table pattern is better

I don't think we need to show diffs, we just need to tell the developer where the duplicate IDs are so that they can fix the problem.

> Figuring out what data to display to the user in order to trace down what workflows are duplicated has proven to be a tricky one

I see... yes good point. That kinda sux. I wonder if there's something we can do (future potential improvement / research) where the `Define*` functions can use the various module metadata APIs deno provides in order to figure out the source location. Maybe each `*Definition` object logs its `import.meta.url` in a private property, the main `Manifest` object does the same, and then some of the `Manifest` methods can use `import.meta.resolve()` between the Manifest's URL and the various `*Definition` URLs to figure out source locations?

WilliamBergamin: > I wonder if there's something we can do (future potential improvement / research)

Did some research on this earlier seems like its might be a complicated solution, I will dumb down this PR and ask for a second round of review

filmaj (approved):
Awesome! :shipit:

PR #194, opened 2023-08-03
Consolidate oauth2 provider type

WilliamBergamin:
###  Summary

This PR aims to address [this comment](https://github.com/slackapi/deno-slack-sdk/pull/192#discussion_r1276486102) that asks to consolidate `ManifestOAuth2ProviderSchema` and `OAuth2ProviderDefinitionArgs`

I've left a TODO comment that mentiones that some work regarding the `type` system could be done once more then one Provider types are present.

#### testing
1. Open an app with an OAuth2 provider such as [deno-simple-survey](https://github.com/slack-samples/deno-simple-survey)
2. replace the `"deno-slack-sdk/"` value with `"https://raw.githubusercontent.com/slackapi/deno-slack-sdk/consolidate-Oauth2ProviderType/src/"`
3. Run the app
4. Everything should work as expected

### Requirements

### DIFF
--- a/src/providers/oauth2/types.ts
+++ b/src/providers/oauth2/types.ts
@@ -4,3 +4,33 @@ export type OAuth2ProviderDefinitionArgs = ManifestOAuth2ProviderSchema & {
   /** A unique name for the provider */
   provider_key: string;
 };
+
+/**
+ * TODO: The type system here could be improved one more then one provider type (CUSTOM) is available
+ * provider_name, authorization_url, token_url, identity_config and authorization_url_extras
+ * are only required for CUSTOM provider types
+ */
+export type OAuth2ProviderOptions = {
+  /** Client id for the provider */
+  client_id: string;
+  /** Scopes for the provider */
+  scope: string[];
+  /** Display name for the provider. Required for CUSTOM provider types. */
+  provider_name?: string;
+  /** Authorization url for the provider. Required for CUSTOM provider types. */
+  authorization_url?: string;
+  /** Token url for the provider. Required for CUSTOM provider types. */
+  token_url?: string;
+  /** Identity configuration for the provider. Required for CUSTOM provider types. */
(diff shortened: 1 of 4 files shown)


src/providers/oauth2/types.ts
```diff
@@ -1,35 +1,36 @@
...
-    [key: string]: string;
-  };
+export type OAuth2ProviderDefinitionArgs = ManifestOAuth2ProviderSchema & {
+  /** A unique name for the provider */
+  provider_key: string;
 };
 
+/**
+ * TODO: The type system here could be improved one more then one provider type (CUSTOM) is available
+ * provider_name, authorization_url, token_url, identity_config and authorization_url_extras
```
WilliamBergamin, line 10: In theory these should be required but I fear that making these fields `required` might break some apps

filmaj: How did you determine that these fields are required? Checked the backend schema or some other way?
Can you confirm that not providing these fields works? I.e. can you deploy an app and pass manifest validation if you remove these fields?

filmaj (commented):
Left some questions!

src/providers/oauth2/types.ts
WilliamBergamin: The comments for each of those field mentioned that they we required for `CUSTOM` provider types, I am assuming they are true

But since those fields are currently `not required` I don't think we can make them `required` without potentially breaking apps

filmaj: "Breaking" from a static-type-check perspective, that is true, however, if the backend requires these fields anyways, then that is much less a concern, I think, because no _real_ app would be able to define a provider without them.

That's why I think we should test this out on a real app 😄

WilliamBergamin:
Converted this PR into `draft` since changes surrounding OAuth providers may be coming and we don't want to paint ourselves into a corner 🎨

PR #198, opened 2023-08-16
adding manifest property for  token url configs

uhunnyslack:
###  Summary

This PR is to add two new fields that helps the connector team unblock creating some more connectors.

1. `use_pkce` : boolean -> this is true for providers that uses PKCE:

2. `token_url_config` : { `use_basic_authentication_scheme_scheme` : boolean} to support providers that use this.

More discussions can be found in our slack instance: 


WilliamBergamin:
@uhunnyslack I was working on #194 that aims to consolidate the provider args options `type` and the `type` found in the `manifest` we decided to put this PR on hold waiting to see if your changes could conflict with mine

Do you know if there will be a new provider other changes coming in the near future and would it be worth it to create a typescript type dedicated to PKCE providers?

src/providers/oauth2/types.ts
```diff
@@ -8,6 +8,10 @@ export type OAuth2ProviderIdentitySchema = {
   };
 };
 
+export type tokenUrlConfigSchema = {
+  "use_basic_authentication_scheme"?: boolean;
```
filmaj, line 12: Let's add a JSdoc here to explain what this is used for.

.slack/apps.json
```diff
@@ -0,0 +1 @@
+{}
```
filmaj, line 1: Please remove this file.

filmaj (changes requested):
Overall LGTM, left some comments on removing the apps.json file and adding more docs.

Let's coordinate in Slack re: releasing this.

PR #203, opened 2023-08-23
Adding pkce support field

uhunnyslack:
###  Summary

This PR is to add a new field that helps the connector team unblock creating some more connectors.

`use_pkce` : boolean -> this is true for providers that uses PKCE:

More discussions can be found in our slack instance.

### Important
Please do not merge it until the web app changes are merged! Thanks!


src/manifest/manifest_test.ts
```diff
@@ -972,6 +972,49 @@ Deno.test("SlackManifest() oauth2 providers get set properly", () => {
...
+    botScopes: [],
+    externalAuthProviders: [Provider1, Provider2],
+  };
+  assertEquals(definition.externalAuthProviders, [Provider1, Provider2]);
+  const Manifest = new SlackManifest(definition);
+  const exportedManifest = Manifest.export();
+
+  assertEquals(exportedManifest.external_auth_providers, {
+    "oauth2": {
+      "test_provider_with_with_pkce_true": Provider1.export(),
```
filmaj, line 1012: I think the only thing this test tests is that the provider key is set properly on the exported manifest. Other than that assertion, this test will always pass, as the manifest values for any oauth provider will always be the output of the provider's `export` method. For example, in the future, if I change the [`Provider.export` method](https://github.com/slackapi/deno-slack-sdk/blob/main/src/providers/oauth2/mod.ts#L28) to `delete` the `pkce` field, this test will still pass.

I think we want to change this test to ensure that the `options.use_pkce` field is set properly on the manifest in both cases: it is `true` in one case and `false` in the other.

filmaj (changes requested):
I think the test needs changing as it will always pass, but other than looks fine to me.

src/manifest/manifest_test.ts
uhunnyslack: I tried with some test mutation and seems like I can make the test fail like this:
Screenshot 2023-08-23 at 1 35 15 PM

uhunnyslack: Do you think this would work? ☝️

filmaj: Right, that fails because the key of Provider2 is being checked against the export of Provider1, so we are checking that two different things are equal.

filmaj: I'd suggest enumerating specifically the fields that you expect in each case. Something like:

```
assertEquals(exportedManifest.external_auth_providers.oauth2.test_provider_with_with_pkce_true.options.use_pkce, true);
assertEquals(exportedManifest.external_auth_providers.oauth2.test_provider_with_with_pkce_false.options.use_pkce, false);
```

uhunnyslack: Thanks for the suggestion! Added the changes ! 🙇‍♀️

uhunnyslack: @filmaj  Please do not merge it until the web app changes are merged! Thanks!

filmaj (approved):

filmaj:
@uhunnyslack I've assigned this PR to you and added the do not merge label to it. Feel free to merge whenever you are ready.

PR #213, opened 2023-09-08
Improve code coverage

WilliamBergamin:
### Summary

This PR aims to provide a more accurate representation of code coverage by excluding the integration tests and rearranging some unit tests to explicitly include all the code that is exposed to the user.

### testing

1. pull this branch
2. run `deno task test` to execute all tests
3. run `deno task coverage` to execute the code coverage

### Special notes

I've added a few more pure uni tests and rearranged some file names and structures

### Requirements 


src/schema/slack/functions/_scripts/src/write_function_files_test.ts
```diff
@@ -0,0 +1,6 @@
+import { assertExists } from "../../../../../dev_deps.ts";
```
WilliamBergamin, line 1: Should we include all the generation script code in our code coverage?
The end user does not directly use this code

filmaj: If we have the tests written already, then sure! It seems like codecov reports uncovered lines in the generation script files, so if we decide to remove, then we should make sure to turn that off, too.

filmaj (approved):
LGTM! But I did have one failure when running on deno v1.36.3 🤔 

```
AddBookmark generates valid FunctionManifest => ./src/schema/slack/functions/add_bookmark_test.ts:9:6
error: AssertionError: Values are not equal:


    [Diff] Actual / Expected


    {
-     description: undefined,
```

src/schema/slack/functions/_scripts/src/write_function_files_test.ts
WilliamBergamin: Unit tests for the generation scripts are currently minimal, I will therefore exclude these files from our code coverage

implementing stable unit tests for this section is complicated since the returned values of the functions are strings of partially formatted typescript


In a follow up PR I've been able to implement an integration test that ensure that the generated code is valid typescript

PR #217, opened 2023-09-15
Adding http post method config

uhunnyslack:
###  Summary

This PR is to add a new field that helps the connector team unblock creating some more connectors.

```
 "body"?: {
    [key: string]: string;
  };
  "http_method_type"?: IdentityUrlHttpMethodTypeValues;
```

More discussions can be found in our slack instance.

### Important
Please do not merge it until the web app changes are merged! Thanks!


src/providers/oauth2/types.ts
```diff
@@ -1,11 +1,23 @@
 import { OAuth2ProviderTypeValues } from "../../schema/providers/oauth2/types.ts";
 
+const IdentityUrlHttpMethodTypes = {
+  POST: "POST",
+  GET: "GET",
+} as const;
+
+export type IdentityUrlHttpMethodTypeValues =
```
WilliamBergamin, line 8: Do we want users to do something like this?

```ts
  const Provider2 = DefineOAuth2Provider({
    provider_key: providerKey2,
    provider_type: Schema.providers.oauth2.CUSTOM,
    options: {
      "client_id": "123.456",
      "scope": ["scope_a", "scope_b"],
      "identity_config": {
        "url": "https://example.com",
        "account_identifier": "account_identifier_string",
        "headers": { "key1": "header_1", "key2": "header_2" },
        "body": { "param1": "body_1", "param2": "body_2" },
        "http_method_type": IdentityUrlHttpMethodTypes.POST,
      },
    },
  });
```

For http method values , it could be reasonable to ask the user to always pass the string value instead of a reference to an object we define (this seems to be the implementation of the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options)) we could keep the `typescript` simple and define the following
```ts
IdentityUrlHttpMethodTypeValues = "GET" | "POST"
```

let me know what you think?

src/manifest/manifest_test.ts
```diff
@@ -1433,3 +1433,120 @@ Deno.test("SlackManifest() oauth2 providers get set properly with token_url_conf
     true,
   );
 });
+
+Deno.test("SlackManifest() oauth2 providers get set properly with identity_config", () => {
```
WilliamBergamin, line 1437: NIT: This file seems large and there may be more tests added to it in the future
Maybe we could create `src/providers/oauth2/mod_test.ts` that would contain tests related to the `src/providers/oauth2/` folder 💡 

Let me know what you think, I am unsure if this will trigger a larger refactor of tests

uhunnyslack:
@WilliamBergamin I have addressed the review comments as suggested.
Please review whenever possible. Thanks!

filmaj (changes requested):
I left a minor request to document the properties, but looks good to me otherwise.

We should also engage the documentation team to update the external-auth docs (https://api.slack.com/automation/external-auth) with these new additions.

uhunnyslack:
> 
`We should also engage the documentation team to update the external-auth docs`

@filmaj 
This is a good point. 
We have separate tickets to track the external documentations. Will be tackling them as soon as possible (after tackling the current priorities)
Thanks!

filmaj (approved):
Thank you and LGTM! Feel free to merge whenever.

WilliamBergamin (approved):
Thanks for the changes 💯

PR #278, opened 2024-02-20
Add time_to_live_attribute to datastore module

mniemer:
### Summary

This PR adds support for the AWS DynamoDB TimeToLive feature on Slack app datastores. This feature allows developers to identify an attribute in their datastore schema which serves as an expiration timestamp. In the background, DynamoDB will check for items which have this attribute populated with an epoch timestamp < now and delete them from the DynamoDB table. 

To enable the feature, add `time_to_live_attribute` in the `DefineDatastore` function arguments and specify an attribute from the `attributes` map which has type `Schema.slack.types.timestamp`.

Example:
```
const SongDatastore = DefineDatastore({
  name: SONG_DATASTORE,
  primary_key: "id",
  time_to_live_attribute: "song_played_date",
  attributes: {
    id: {
      type: Schema.types.string,
    },
   ...
    song_played_date: {
      type: Schema.slack.types.timestamp,
    },
  },
});
```

Inclusion of the `time_to_live_attribute` is optional, and if it is not included, the time to live feature will not be enabled for the datastore.

### Testing

Manual tests:
- Including `time_to_live_attribute` leads to enablement of TTL feature for that attribute
- Not including `time_to_live_attribute` leads to disablement of TTL if it was previously enabled
- Validations are enforced by Slack APIs `apps.manifest.create` & `apps.manifest.update`

### Special notes

- This feature also requires a change in slack-cli to propagate the new `time_to_live_attribute` to the Slack API: https://github.com/slackapi/slack-cli/pull/1252
- On Slack server side, the feature is gated by a feature flag which is enabled for internal Slack test teams. If the feature flag is not enabled, the inclusion of `time_to_live_attribute` will be ignored by the Slack APIs `apps.manifest.create` & `apps.manifest.update`

### Requirements

### DIFF
--- a/src/manifest/manifest_schema.ts
+++ b/src/manifest/manifest_schema.ts
@@ -273,7 +273,7 @@ export type ManifestCustomTypesSchema = {
 export type ManifestDatastore = ISlackDatastore;
 export type ManifestDatastoreSchema = {
   primary_key: string;
-  time_to_live_attribute: string;
+  time_to_live_attribute?: string;
   attributes: {
     [key: string]: {
       type: string | ICustomType;
--- a/src/manifest/manifest_schema.ts
+++ b/src/manifest/manifest_schema.ts
@@ -273,6 +273,7 @@ export type ManifestCustomTypesSchema = {
 export type ManifestDatastore = ISlackDatastore;
 export type ManifestDatastoreSchema = {
   primary_key: string;
+  time_to_live_attribute: string;
   attributes: {
     [key: string]: {
       type: string | ICustomType;
--- a/src/datastore/mod.ts
+++ b/src/datastore/mod.ts
@@ -16,8 +16,14 @@ export const DefineDatastore = <
   Name extends string,
   Attributes extends SlackDatastoreAttributes,
   PrimaryKey extends keyof Attributes,
+  TimeToLiveAttribute extends keyof Attributes,
 >(
-  definition: SlackDatastoreDefinition<Name, Attributes, PrimaryKey>,
+  definition: SlackDatastoreDefinition<
+    Name,
+    Attributes,
+    PrimaryKey,
+    TimeToLiveAttribute
+  >,
 ) => {
   return new SlackDatastore(definition);
 };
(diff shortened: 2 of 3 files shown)


src/manifest/manifest_schema.ts
```diff
@@ -273,6 +273,7 @@ export type ManifestCustomTypesSchema = {
 export type ManifestDatastore = ISlackDatastore;
 export type ManifestDatastoreSchema = {
   primary_key: string;
+  time_to_live_attribute: string;
```
WilliamBergamin, line 276: Shouldn't this be optional? 🤔

WilliamBergamin (changes requested):
This is awesome 💯  thanks for the contribution

What happens if a user set an attribute as `time_to_live_attribute` and does not give it a value at runtime here is my example 

```ts
export default DefineDatastore({
  name: "issue_datastore",
  primary_key: "id",
  time_to_live_attribute: "issue_expire_date",
  attributes: {
    id: {
      type: Schema.types.string,
    },
    issue_expire_date: {
      type: Schema.slack.types.timestamp,
    },
  },
});

...
    const putResponse = await client.apps.datastore.put<
      typeof IssueDatastore.definition
    >({
      datastore: IssueDatastore.name,
      item: { id: crypto.randomUUID() },
    });
```
Should the attribute selected as being the `time_to_live_attribute` be set as a required field in typescript similar to how the primary key made required?

mniemer:
> What happens if a user set an attribute as time_to_live_attribute and does not give it a value at runtime... Should the attribute selected as being the `time_to_live_attribute` be set as a required field in typescript similar to how the primary key made required?

In this case, the item without the TTL attribute will not be eligible for deletion by Dynamo. I think this is a valid usage of the feature - in my own demo I've created a daily song recommender that uses the attribute `song_played_date` as the TTL attribute:
```
const SongDatastore = DefineDatastore({
  name: SONG_DATASTORE,
  primary_key: "id",
  time_to_live_attribute: "song_played_date",
  attributes: {
    id: {
      type: Schema.types.string,
    },
...
    song_played_date: {
      type: Schema.slack.types.timestamp,
    },
  },
});
```

I have two custom functions, one where users of the app submit songs, and another where the app picks a song from the suggestion list to recommend to the users. The `song_played_date` attribute is only added once the song is recommended by the second function, marking it as ready for expiration:

```
    // Add song_played_date to song datastore
    const played_date = Math.trunc(Date.now().valueOf() / 1000);
    const songPutResponse = await client.apps.datastore.put({
      datastore: SONG_DATASTORE,
      item: {
        id: song.id,
...
        song_played_date: played_date,
      },
    });
```

WilliamBergamin:
Ohh I see, this is a great implementation 💯  if no value is passed to the attribute selected as the `time_to_live_attribute` then the entry is not eligible for deletion

WilliamBergamin (approved):
LGTM 💯  well thought out solution, but lets wait for feedback on the CLI changes before merging this to `main`

PR #302, opened 2024-04-15
Updates to built in funcs based on latest `functions.list` output

### LINKED ISSUE #267: [BUG] Unable to pass `team_id` to built-in Slack function `Schema.slack.functions.CreateChannel`
**Question**

I'm trying to use `Schema.slack.functions.CreateChannel` in a new Deno workflow, and I want to know how I should pass the team_id parameter. It's something similar to how it's done with the Slack API client.

API client
```
client.conversations.create({
         name: "some-name",
         is_private: false,
         team_id: "myteamid",
});
```

Slack function CreateChannel as workflow steps does not accept team_id
```
createIncidentWorkflow.addStep(
  Schema.slack.functions.CreateChannel,
  {
    channel_name: "some-name",
    is_private: false,
    // team_id: "myteamid" <== this breaks the workflow
  },
)
```

**Context**

`slack run` works as expected, but when trying to create a channel, my workflow gets stuck. I believe it's due to restricted workspace permissions. I'm working on a "development" workspace, where I'm admin, so I want to make sure the function above creates a channel in the correct workspace. 

**Environment**

```
cat import_map.json | grep deno-slack
    "deno-slack-sdk/": "https://deno.land/x/deno_slack_sdk@2.5.0/",
    "deno-slack-api/": "https://deno.land/x/deno_slack_api@2.1.2/",
    "deno-slack-hub/": "https://deno.land/x/deno_slack_hub@0.1.1/",
```

```
deno --version
  deno 1.33.3 (release, aarch64-apple-darwin)
  v8 11.4.183.2
  typescript 5.0.4
```

```
sw_vers && uname -v
ProductName:            macOS
ProductVersion:         13.6.3
BuildVersion:           22G436
Darwin Kernel Version 22.6.0: Tue Nov  7 21:40:08 PST 2023; root:xnu-8796.141.3.702.9~2/RELEASE_ARM64_T6000
```

filmaj:
I don't think we want to merge this as it has some breaking changes and other weird bits. Specifically:

- `canvas_copy` is now removed? it is publicly documented and I imagine this is a breaking change.
- `create_channel` and `create_usergroup` now accepts a `team_id` property (this is good; would resolve #267), but their type is undefined?
- `openform` has a new parameter that is maybe supposed to be hidden? "Workflow step configuration for a function to be executed on submission of the form"


filmaj:
Closed in favour of #361

PR #311, opened 2024-05-02
Add Share Canvas in thread to the deno sdk

rafaelamsili:
### Summary

This PR add the new released Share Canvas in thread to the deno sdk and update labels for the owner in copy canvas, create canvas and create channel canvas

| Field  | Type | Description | Required
| -------- | ------- | ------- | ------- |
| canvas_id  | slack#/types/canvas_id | The id of the canvas that will be shared    | true |
| message_context |  slack#/types/message_context | The reference to the root message that the canvas will be shared as a reply    |  true |
| access_level | string | The access level that the canvas will be shared. It can be `view` or `edit`   |  true  |
| message    | slack#/types/rich_text | Optional rich text message that will be send when sharing the canvas  |  false |
| reply_broadcast    | boolean | If selected will also broadcast the message to the channel |  false - defaults to `false` |

### Testing

Create an app using the functions and see that it perform the actions

### Requirements 
- Canvas need to be enabled in the org


src/schema/slack/functions/canvas_copy.ts
```diff
@@ -22,7 +22,7 @@ export default DefineFunction({
       },
       owner_id: {
         type: SlackTypes.user_id,
-        description: "Canvas owner id",
+        description: "Person",
```
rafaelamsili, line 25: This label changed in the built in

filmaj:
@rafaelamsili did you run the generation script for this (it seems to me like no)? Because the `mod.ts` is missing the import of the new function, so there's no way for a developer to actually use this as-is.

filmaj (changes requested):
The one thing we're missing is importing the new function in this file and exposing it as part of the export: https://github.com/slackapi/deno-slack-sdk/blob/main/src/schema/slack/functions/mod.ts

filmaj (approved):

rafaelamsili:
> @rafaelamsili did you run the generation script for this (it seems to me like no)? Because the `mod.ts` is missing the import of the new function, so there's no way for a developer to actually use this as-is.

Good catch 🦅 👁️  about the mod change. I ran the generation script but there was a lot of noise from other functions so I just committed those changes.

PR #327, opened 2024-06-06
New `team_id` type, updates to built-ins, updated builtin function generation script instructions

filmaj:
- Adds new Add Bookmark built-in function.
- Adds a new `team_id` type
- Adds a `team_id` parameter to Create Channel and Create Usergroup builtins (relates to but does not resolve it #267)
- Updates to the generation script and instructions:
  - For the foreseeable future, the `functions.list` API will be in flux as it is under active development. As such, the generation instructions have changed to use a different set of APIs as a replacement. The new way is a bit more involved manual process, but it is not so bad (7 API calls using the API Tester instead of 1).
  - The generation script has unfortunately to this point ignored the `is_hidden` parameter. This means we have released hidden parameters in this SDK 😬 to maintain backwards compatibility, handling of hidden parameters is now a manual process. The generation script will warn if it encounters a hidden parameter. Instructions are updated to have the maintainer carefully review the generated code with respect to hidden parameters. Existing hidden parameters will be kept, but any future hidden parameters should be removed.

# How To Deal With Hidden Parameters

We could instead improve the generation script to manually hard-code the list of built-in functions that have had hidden parameters previously exposed, and ignore those / proceed with generation code for them (to maintain backwards compatibility). Any _other_ hidden parameters _except_ the ones already exposed will be skipped for code generation. Thoughts @WilliamBergamin ?


deno.jsonc
```diff
@@ -1,35 +1,29 @@
...
-    },
-    "options": {
-      "semiColons": true,
-      "indentWidth": 2,
-      "lineWidth": 80,
-      "proseWrap": "always",
-      "singleQuote": false,
-      "useTabs": false
-    }
+    "include": [
```
filmaj, line 4: Flattening out `deno.jsonc` to remove deprecation warnings from `deno` output

src/schema/slack/functions/add_bookmark.ts
```diff
@@ -0,0 +1,50 @@
+/** This file was autogenerated. Follow the steps in src/schema/slack/functions/_scripts/README.md to rebuild **/
```
filmaj, line 1: New built-in that's been public for a little bit.

src/schema/slack/functions/create_channel.ts
```diff
@@ -9,6 +9,11 @@ export default DefineFunction({
   title: "Create a channel",
   input_parameters: {
     properties: {
+      team_id: {
```
filmaj, line 12: This is the new property that fixes #267

src/schema/slack/types/mod.ts
```diff
@@ -1,17 +1,18 @@
 const SlackPrimitiveTypes = {
-  user_id: "slack#/types/user_id",
+  blocks: "slack#/types/blocks",
```
filmaj, line 2: Just alphabetizing these

src/schema/slack/types/mod.ts
```diff
@@ -1,17 +1,18 @@
...
   expanded_rich_text: "slack#/types/expanded_rich_text",
-  rich_text: "slack#/types/rich_text",
-  message_ts: "slack#/types/message_ts",
   file_id: "slack#/types/file_id",
-  canvas_id: "slack#/types/canvas_id",
-  canvas_template_id: "slack#/types/canvas_template_id",
+  message_ts: "slack#/types/message_ts",
+  oauth2: "slack#/types/credential/oauth2",
+  rich_text: "slack#/types/rich_text",
+  team_id: "slack#/types/team_id",
```
filmaj, line 12: New type here; I am about to test this manually in an app to make sure it works.

WilliamBergamin (approved):

PR #369, opened 2024-10-21
E2E Test System integration

filmaj:
Adds a CircleCI config to kick off end-to-end platform integration tests and report results back, passing in the branch ref of the deno-slack-sdk as a parameter into the tests.

This should execute on both PRs as well as on `main`, and pass the correct branch in either case.


.circleci/config.yml
```diff
@@ -0,0 +1,70 @@
...
+            -u "${FIL_CIRCLECI_PERSONAL_TOKEN}:" \
+            --data "{\"branch\":\"<< parameters.e2e_target_branch >>\",\"parameters\":{\"deno_sdk_import_url\":\"${IMPORT_URL}\"}}" | jq '.id')
+            echo "e2e test workflow started with id: $TEST_JOB_WORKFLOW_ID"
+            echo "export TEST_JOB_WORKFLOW_ID=${TEST_JOB_WORKFLOW_ID}" >> $BASH_ENV
+      - run:
+          name: Wait for platform-devxp-test E2E run to complete
+          command: |
+            E2E_RESULT="{}"
+            E2E_STATUS="running"
+            while [[ $E2E_STATUS != "failed" && $E2E_STATUS != "canceled" && $E2E_STATUS != "success" ]]
```
WilliamBergamin, line 42: What are the other states 🤔 
Any risk of an endless loop? or waiting for a timeout when the process could fail fast

.circleci/config.yml
```diff
@@ -0,0 +1,70 @@
...
+            do
+              sleep 10
```
WilliamBergamin, line 44: What is the reasoning behind `10`?

Also adding a [specifier suffix](https://www.cyberciti.biz/faq/linux-unix-sleep-bash-scripting/) can make it clearer for new bash devs
```suggestion
              sleep 10s
```

WilliamBergamin (changes requested):
Not an expert in this but it looks good 💯 

left 2 comments

.circleci/config.yml
filmaj: Wow didn't know you could do that! Great suggestion. As for 10 seconds: no specific reason. Our tests take a minimum of 1min to run. We could probably bump this delay up to 30s without a huge impact to us and be slightly nicer to our friends at CircleCI.

WilliamBergamin (approved):
:shipit:

